PR #5, opened 2018-09-13
Use MatchPy to compile MoA

saulshanabrook:
In this PR I am working on compiling array expressions, including MoA expressions. I want to be able to express the MoA equivalences of indexing and shapes for different operations and to build up an expression of them and then execute that on Python backed arrays (I am using tuples right now) and NumPy backed arrays.

We should also be able to define NumPy operations that can either be compiled to MoA expressions or executed directly on NumPy arrays.

I am using [MatchPy](https://matchpy.readthedocs.io/en/latest/) to define replacement expressions and reduce them. 

My MVP is inner and outer product and NumPy equivalences, see https://medium.com/@saulshanabrook/matrix-products-mathematics-and-implementations-4d13b7112f94 for math.


saulshanabrook:
This is working enough for cross products, so I am merging this in.

PR #33, opened 2018-11-04
Start adding docs to readme

### LINKED ISSUE #25: Remove `value` of Value type
Matchpy sets default `name` attribute we should use

saulshanabrook:
Thanks to a helpful discussion with @dcharbon on Friday, I have started outlining how this system works. 

- [x] Finish writing
- [ ] Think about if we should change names to make them more math (category theory)-ey. 
- [x] Add `replace_scan`
- [x] Test readme code
- [ ] Make notebook instead

Fixes #25


saulshanabrook:
In this process, I realize I can add typing and register an operation very cleanly as a function!

Going from this:

```
class GetItem(matchpy.Operation):
    name = "GetItem"
    arity = matchpy.Arity(1, True)

# then in .pyi file
def GetItem(seq: CArray) -> CGetitem:
    ...
```

To this:

```python
@operation
def GetItem(length: CContent, getitem: CGetitem) -> CArray:
    ...
```

And `operation is pretty simple as well:

```
T = typing.TypeVar("T", bound=typing.Callable)


def operation(fn: T) -> T:
    """
    Register a matchpy operation for a function. 

    The function should have a fixed number of args and one return.

    This can also be done manually, by typing the result of `matchpy.Operation.new`
    as a callable, but this is more fluent.
    """
    sig = inspect.signature(fn)
    return matchpy.Operation.new(fn.__name__, matchpy.Arity(len(sig.parameters), True))
```

saulshanabrook:
Trick is understanding that all this is just for compile time checking, not runtime.

saulshanabrook:
OK this should be ready to merge. I have added some tests, but no CI yet. The tests just run some notebooks, run mypy and verify the readme works.

saulshanabrook:
Closes #25

saulshanabrook:
CI is working now on this and it is ready to merge.

uarray/uarray/core.py
```diff
@@ -1,281 +1,280 @@
...
-register(Length(Sequence(w.length, w._)), lambda _, length: length)
+class CCallableBinary(typing.Generic[RET, ARG1, ARG2]):
+    pass
 
 
-class Scalar(matchpy.Operation):
-    """
-    Scalar(content)
-    """
+CGetitem = CCallableUnary[CArray, CContent]
```
hameerabbasi: Is it possible to rename this to `CGetItem`?

saulshanabrook: Yes :) Thanks for catching this.

saulshanabrook:
@hameerabbasi good idea, I moved the types into separate files. How is that?

hameerabbasi:
It's better, I was also wondering about the naming convention, what does the `C` signify? Is it necessary to have it? Python just has `PascalCase` convention, usually. Are you following the C++ way of prefacing type names with `C`?

It'd be nice to have all of them in a simple `typing` module, and we can make the rest of them private. Another way is to have them close to where they should be.

Examples of both: `typing.List` and `collections.Iterable`.

saulshanabrook:
C stands for Category. All the names will need to changed, i have gotten feedback from a couple people that the resemblance to Pythons naming (GetItem, Sequence) is confusing. 

At least at the moment within this PR it is internally consistent that all compile time types are prefixed with C.

I expect the final look of how we write these to change. 

I'll add a few lines to the readme to explain how they are currently.

hameerabbasi:
Perhaps have a `Category` base class? It could come in useful either way.

saulshanabrook:
Sounds good. Added that.

saulshanabrook:
I am gonna merge this in so that I can share the current state of the repo with others for feedback more easily.

PR #41, opened 2018-11-14
Add more numpy examples

saulshanabrook:


saulshanabrook:
OK this is ready to go now! We have reshaping working and broadcasting is fixed

PR #67, opened 2018-11-26
Adding initial conda build recipe

costrouc:
This is a simple pull requests to at least add a build recipe for conda. Currently the recipe uses `v0.4` which has `pyproject.toml` in the `uarray` directory. Things that need to be addressed.

 - `.conda` a good directory to store the recipe?
 - should releases be integrated into `azure-pipelines.yml`?  *edit* upon looking further is looks like azure pipelines [do not support releases yet](https://stackoverflow.com/questions/52323065/azure-devops-yaml-release-pipelines)


costrouc:
Will do. I will add short documentation for `flit` pypi release along with `conda`. In order to add a shield and conda installation instructions a release will need to be made. I have a registry [here](https://anaconda.org/costrouc) would it be okay for now to publish it there?

Installation instructions will look like `conda install -c costrouc uarray`.

saulshanabrook:
Great, thank you.

Sure that registry would be fine. Or you could make one called like `uarray`, so that others could be added to it.

saulshanabrook:
This looks good. Can I merge?

costrouc:
Right now the conda image only works with `linux`. I am going to add the `noarch` flag so the package works on windows, linux, and OSX. Should be ready in 15 min

costrouc:
I have tested that `uarray` conda installation works within a docker image with a `noarch` package. So ready for merge. I have also added conda installation instructions.

Issue #65, opened 2018-11-19
Switch Vector to nested type


saulshanabrook:
Like a recursive Cons list

saulshanabrook:
This would unify `PushVector` and `Vector` representations.

Issue #52, opened 2018-11-18
Multiply internal representations: N-dimensional, vectors, Nested sequences


saulshanabrook:
Notes from DB talk:

> How do schemas fit in our world? 
> 
> This is what dimensionality and shape information are. They are different levels of schemas. 
> 
> Need way to talk about all these at the same time and convert between them. Simply! 
> 
> Nested sequence vs ND array. They are just different schemas. 
> 
> Need to validate our specs. Generate random arrays. Need specs for types! Esp integer. Integer ranges...

Sequences are the most expressive, but we need to be able to think about arrays in other forms. Here are some examples of different interfaces in Python land:

* N-dimensional arrays: nested-sequence is superset of this. have dim, shape, and indexing: Numpy arrays
* Nested Sequences:  Recursive indexing. nd array can translate to this. Lists of lists
* Vectors: Any indexing to value, contiguous memory. N-Dimensional and Nested Sequences can translate to this. lists, JS arraybuffer

We want to be able to write algorithms that target any of these and we can have translations between them. i.e. any algorithms written for nested sequences can be used on n-dimensional arrays, b/c nd arrays can always be translated to nested sequences. Reverse is not true, not all nested sequences can be translated to arrays (if ragged).

We can always map from ND arrays to vectors, by using gamma functions and reverse using inverse gamma and shape.

To translate from nested sequences to flat vectors requires also making some offset vector.

PR #73, opened 2018-12-05
Switch to ndarray from nested arrays

### LINKED ISSUE #35: Look into fixed point forms
After reading http://conal.net/papers/essence-of-ad/ I wonder if we can convert our functional forms into fixed point. aka instead of having lambda functions turn them all into composition and other higher level forms.

Why? Not sure, does this get us AD? Need to review that paper more to understand crux of requirements.

saulshanabrook:
This PR attempts to make the codebase simpler to reason about.

- [x] Switches everything to be rectangular N-dimensional arrays instead of nested sequences
  - [x] add dimensionality to arrays
  - [ ] Add natural numbers with range
  - [x] switch everything to category theory language, to differentiate from python concepts (callable)
- [x] Type all replacements (this also moves us from writing farther from matchpy implementation, so we could more easily switch to different backend if need be)
- [ ] Switch AST generation to convert from nested function forms
- [ ] Convert MoA expressions to use full form

Will close #35 #42 #52 #55 #65


saulshanabrook:
@costrouc This is ready to merge from my end!

The code generation and numpy integration isn't working anymore, but everything has a much more solid foundation and they can be added in on top of this.

Check out this notebook for the paper problem worked out: https://github.com/Quansight-Labs/uarray/blob/a82225c253b10ea3022c639ee8e95be60b48428a/notebooks/Paper%20Problem.ipynb

saulshanabrook:
Closing in favor of #85

PR #74, opened 2018-12-06
Adding initial MOA parser

costrouc:
@saulshanabrook  what are your thoughts?

- tests are included with parser

It can handle pretty much any moa expression. There are still a few symbols that need to be added but the fundamentals are covered. Additionally it will complain when it does not understand an expression.

Future vision of expressions. 
 - We need an ability for composability. For instance `<1 2> psi A` can be given an name like `my_index` and can be used in a later expression like `my_index * B`.
 - Supplying variables can be done with a dictionary like so `moa.eval('<1 2> psi A', {'A': np.random.random((2, 3, 4)))`

Currently pretty much any moa expression will evaluation but it will complain that uarray does not have support for the operation such as ('/' and '-').

Examples. Available tokens can be found in `lexer.py`.

```
<1 2> + <3 4>

A + B + C + D + E

A + (B + C) + D psi E

A ^ <1 2> + B psi C
```


saulshanabrook:
I think you should be able to use `vector` from `core.py`: https://github.com/Quansight-Labs/uarray/blob/c1542e471d9302dcdbcbf7cc97e50ae5d5dd815f/uarray/core.py#L238-L243

and `with_shape` from core as well. I think you could do something like:

```python
with_shape(unbound("A"), [Int(1), Int(2)])
```

To get an an array named "A" with shape `<1 2>`.

I would continue to based off of core. My PR is in a lot of flux and we should be able to move this over easily if we do end up merging it eventually.

costrouc:
I will add a notebook to demonstrate the features! Perfect you gave me exactly what I needed to know.

costrouc:
Tests fail due to the following not being equal.

```
 A = core.with_shape(core.unbound('j'), [moa.Int(1), moa.Int(2)])
    B = core.with_shape(core.unbound('j'), [moa.Int(1), moa.Int(2)])
    print(A)
    print(B)
    assert A == B
```

```
Sequence(1, (i6 -> Sequence(2, (i7 -> Scalar(Content(CallUnary(GetItem(CallUnary(GetItem(j), i6)), i7)))))))
Sequence(1, (i8 -> Sequence(2, (i9 -> Scalar(Content(CallUnary(GetItem(CallUnary(GetItem(j), i8)), i9)))))))
```

saulshanabrook:
@costrouc I would do something like:

```python
 A = core.with_shape(core.unbound('j'), [moa.Int(1), moa.Int(2)])
 B = core.with_shape(core.unbound('j'), [moa.Int(1), moa.Int(2)])

 assert replace(Shape(A)) == replace(Shape(B))
idxs = vector_of(unbound("index_1"), unbound("index_2"))
assert replace(Index(idxs, A)) == replace(Index(idxs, B))

```

saulshanabrook:
To make sure that shape is equal and indexed forms are equal

saulshanabrook:
Also, you might not need assertions for now. Even a notebook just showing off this functionality with no tests is useful!

costrouc:
@saulshanabrook this is what I was thinking. For now I have the specific tests that test `A ^ <1 2>` logic disabled. Everything else works. I am developing notebooks now.

costrouc:
Added notebook and ready for merge. I think that in combination with the visualization writing expressions should be much easier

saulshanabrook:
Wow GIthub has a merge conflict UI now! Pretty sweet.

costrouc:
Yeah it does! But be careful. At least from my experience I've had it duplicate code when I though I was editing a single line.

PR #75, opened 2018-12-06
Adding visualization to uarray expressions

### LINKED ISSUE #40: Add visualizations with graphviz
Use https://graphviz.readthedocs.io/en/stable/examples.html#btree-py to visualize tree. Per operation, make sub-text be name for that argument

costrouc:
Looks like `matchpy` does not want to upstream visualization which is understandable. https://github.com/HPAC/matchpy/pull/48

For now this PR adds a function `visualize_expression` which will return a graphviz object that will easily render in jupyter notebooks. This PR will require additional work in that we would like to add a notebook demonstrating the visualizations. 

Action is needed on which Symbols/Objects in uarray to give the `_repr_gviz_node_` to which will allow for gviz visualization. For now it has a default representation for Symbols/Operations.

Closes #40


costrouc:
Need to add a notebook to demonstrate visualization.

costrouc:
Notebook now exists to demonstrate. Here is a screenshot of visualizing `A^< 4 5 > + B + <1 2> phi C`

screenshot-localhost-8888-2018 12 07-16-25-39

`visualize_expression(moa_expr, with_uarray=False)`

screenshot-localhost-8888-2018 12 07-16-28-22

So both views are possible. I think that the first is more helpful for viewing reductions.

@saulshanabrook if this is good with you I will merge. I know that I can merge anytime but for now I would like to ask for approval first.

PR #85, opened 2019-01-18
Typed Graph Implementation

saulshanabrook:
## TODO
- [x] Get full vertical stack working
  -  [x] Add compilation to Python functions
   - [x] Add compilation to Python AST
   - [x] ~~Add compilation to Tensorflow Graph (Waiting on 3.7 support for tensorflow~~
   - [x] ~~Add compilation to LLVM ~~
- [x] Add different "create" method
- [x] Split MoA into it's own context
- [x] Add NumPy wrapper
  - [x] Get enough to run some benchmarks
- [x] Add sense of "concreteness" to all values. So that default abstractions still only match on "concrete" types. This allows enough staging to wait till something is discrete before deciding how to dispatch on it (general vs. specific).
  - [x] This requires adding a new core thing, like operations. But then we need way to map over children, for replacement. This is like copy functionality. So let's add a mapping function.
- [x] Move MoA to class
- [x] Change ndarray to be array
- [x] Change array idx function to actually take list, but allow creation based on taking vector  (so that we know length of list if we know shape)
- [x] Change all lists to be abstractions really
- [x] Figure out if why we need separate "Data" dataclass, if not, convert everything to operations and type them as always having functions as arguments.
- [x] Change imports so that MyPy can pass on it
- [x] Formalize that operation always has a function as a name, and rename it to 
- [x] Remove data
- [x] Formalize `isinstance Operation` pattern, by adding getter that raises exception.
- [x] Rename things
  - [x] Nat -> Natural
  - [x] ~~Abstraction -> Callable~~
  - [x] ~~"Value" -> Term~~
- [x] Change registration to take two arguments, two converter function, and then next argument if default registered. Both of same type.
- [x] Change all box arguments to have defaults.
- [x] Change NumPy AST output to use ravel and then to iterate
  - [x] Reshape resulting NumPy array
- [ ] Change default display to be `display_ops` so it is more usable
- [x] Rename `_replace` to `replace`
- [x] Remove `import *`, make all imports absolute
  - [ ] Keep data classes as small as possible. Move constructors to other classes?
- [x] Verify paper example and other example 
- [ ] Update README, remove old content
  - [x] Add list of high level concepts
  - [ ] Create examples of different ways of "using" numpy arrays at compile time.
  - [ ] Add multiple narrative notebooks to introduce concepts
     - [ ] Basic types/replacements
     - [ ] Arrays / operations on arrays
     - [ ] Compiling to different backends
  - [ ] Add design intentions
    - [ ] Cartesian closed categories terms / typed lambda calculus correspondence
    - [ ] Python type system, classes vs. functions, fundamental problem of making application lazy
    - [ ] Keeping things as MyPy checkable as possible -> not using fancy python constructs
   
- [x] Allow global context to be editable. 
- [ ] Change to be immutable
- [ ] Make abstractions replace in one go, to simplify reduction graphs.
- [ ] Add multi arity abstractions
- [ ] Figure out how to get `extract_operation` to type properly 

---
Later date:
- [ ] Detach types from values. Figure out how to do this type wise, how not to include values inside of types.
- [ ] Add create methods on array when you know just know name and shape.
- [ ] Make sure args are concrete before turning to AST (do all addition first)
  - [ ] To do this, we might have to implement casting. i.e. if you do (+ 1 AST(...)) we should know that we should convert 1 to an ast first, then use the AST definition of addition. Why? Otherwise how do we convert `(to_ast (+ 1 (loop ...))`? How do we know if we should push `to_ast` inside or wait for inside to resolve?
- [ ] Compute property dtype for numpy
- [x] ~~Understand how AST creation / boxing/operations relates to monads or other category theoretic concepts~~
- [x] ~~Make children and values protocol instead of single dispatch (optional)~~
- [ ] See if we can get away with making all replacements just abstractions. Understand reasoning why we need two systems here (i.e. make meta system be implemented as abstractions)


saulshanabrook:
TODO later:

- [ ] Make lists just abstractions

saulshanabrook:
I would like to merge this in, so we at least at where we were in master in uarray.

Issue #93, opened 2019-02-12
Separate into multiple packages


saulshanabrook:
Here is a sketch of how we could separate the python packages:

* `uarray`: Core types and dispatching infrastructure
* `umoa`: MoA operations and reductionson them.
* `unumpy`:
  * `ndarray.py`: Lazy numpy ndarray that creates graph. Use `__array_ufunc__` and `__array_function__`
  * `ast.py`: Compile to Python AST of numpy operations. Have default implementations use existing numpy functions
* `unumpy_moa`: Adds `moa` translations for a numpy operations, so we can reduce them (remove intermediates, etc)
* `unumpy_llvm`: Adds `llvmlite` compilations for numpy expressions
* `uscipy_linalg`/ `unumpy_linalg`: Adds scipy or numpy compatible linalg module. Depends on `unumpy` and adds default Python AST versions of linalg functions
* `uscipy_linalg_moa`/ `unumpy_linalg_moa`: Adds moa reductions for linalg

MVP could be, AST creation of a bunch of numpy code, where just one bit is fed through MoA  reductions, rest is compiled to existing numpy operations.

Next one would be same, but in LLVM. This would (I think) be more work.

saulshanabrook:
If we want to have a low level representation of operations, one question is should we try to re-use existing SciPy/NumPy implementations? 

[This](https://github.com/numpy/numpy/blob/1c4ab89870e39be7c4a47258f6d92f532650e5cc/numpy/fft/pocketfft.c#L2229-L2285) seems to be an FFT implementation in NumPy. 

It takes in some `PyObject`, gets a `PyArrayObject` from it, then creates some C struct around the FFT options, and calls `rfft_forward` in a loop. 

The interface here seems to be most obviously `PyObject`, if we wanna re-use this code exactly as is. 

Alternatively, we could write our own wrappers around `rfft_forward` I suppose and pass in the data. If we did this, what would our compatibility layer be? 

i.e. we would need to wrap `rfft_forward` in some function. What should the ABI of the function be?

This is where I *want* to reach for gumath, in that it defines an ABI for array functions. But we need to able to compose those (pretty much arbitrarily, create LLVM code around them). And we can't call an gumath kernel inside of a gumath kernel. So this really limits composition.

teoliphant:
I like this plan.  I think we will need to re-wrap the low-level routines (and modifications to cython and f2py can help us do it).

skrah:
[...]

PR #98, opened 2019-02-22
Split into separate Python packages

saulshanabrook:
This splits the `uarray` package into multiple different packages.

I tried to not make any other changes besides moving the Python files around. You should now be able to install them separately and each have their own dependencies.

Fixes https://github.com/Quansight-Labs/uarray/issues/95 and addresses https://github.com/Quansight-Labs/uarray/issues/93

As you can see, in this PR the `unumpy` package depends on the `umoa` package. In the future, we should extract out the MoA implementation of NumPy into a `unumpy_moa` package so that the `unumpy` package does not depend on it. However, I decided to do this work in another PR.


costrouc:
Personal preference but I think it would be cleaner to move the `pyproject.toml` files to each project directory. This would require an additional folder in each project directory.  Otherwise this PR looks good to me.

saulshanabrook:
> Personal preference but I think it would be cleaner to move the `pyproject.toml` files to each project directory. This would require an additional folder in each project directory. Otherwise this PR looks good to me.

Yeah I had it that way originally, then made it less nested. Let's merge this for now, and then we can always switch this if we hit problems down the road.

Thanks for looking it over!

PR #110, opened 2019-03-23
Add CONTRIBUTING.md

jrbourbeau:
This PR adds instructions to `README.md` on how to set up the dev environment and run tests for `uarray`

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,29 +67,6 @@ or
 conda install -c conda-forge -c uarray uarray
 ```
 
+## Contributing
 
-## Development
-
-Make a fork of the main [uarray repository](https://github.com/Quansight-Labs/uarray) and clone the fork:
-
-```
-git clone https://github.com/<your-github-username>/uarray
-```
-
-### Install
-
-`uarray` and all development dependencies can be installed via:
-
-```
-pip install -e ".[all]"
-```
-
-Note that uarray supports Python versions >= 3.5.
-
-### Testing
-
-Tests can be run from the main uarray directory as follows:
-
-```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -61,10 +61,39 @@ This project is in active development and not ready for production use. However,
...
+
+### Install
+
+`uarray` and all development dependencies can be installed via:
+
+```
+pip install -e ".[all]"
+```
+
+Note that uarray supports Python versions >= 3.5.
```
hameerabbasi: We need to tell people here that it currently requires `contextvars` to be installed manually for Python < 3.7.

README.md
```diff
@@ -61,10 +61,39 @@ This project is in active development and not ready for production use. However,
...
+
+### Testing
+
+Tests can be run from the main uarray directory as follows:
+
+```
+pytest
+```
+
 ### Releases
```
hameerabbasi: This whole releases section is invalid now. 🙁

README.md
hameerabbasi: I pushed to `master` so that this isn't required anymore.

hameerabbasi:
I think this should all be in a `CONTRIBUTING.md` rather than the `README.md`.

README.md
jrbourbeau: I'm not sure how releases are currently handled. I can remove this section altogether if you'd like.

hameerabbasi: That'd be nice, I use `twine`. See https://github.com/pydata/sparse/blob/master/release-procedure.md

PR #120, opened 2019-04-01
Use conda task.

hameerabbasi:


xmnlab:
Hey @hammerabbasi  any reason to use conda task? Because it is deprecated now. As you can see in the docstrings here https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/conda-environment?view=azure-devops#yaml-snippet

hameerabbasi:
I reverted that commit.

xmnlab:
:)

PR #125, opened 2019-04-10
WIP: Push initial version of documentation.

hameerabbasi:


docs/index.rst
```diff
@@ -3,8 +3,128 @@
...
+^^^^^^^^^^
+
+``uarray`` (and its sister modules ``unumpy`` and others to come) were designed
+from the ground-up to be modular. This is part of why ``uarray`` itself holds
+the core backend and dispatch machinery, and ``unumpy`` holds the actual
+multimethods. Also, ``unumpy`` can be developed completely separately to
+``uarray``, although the ideal place to have it would be NumPy itself.
+
+However, the benefit to having it separate is that it could span multiple
+NumPy versions, even before NEP-18 (or even NEP-13) was available.
```
rgommers: Also, it can have a faster release cycle which may be useful in the first 1-2 years.

docs/index.rst
```diff
@@ -3,8 +3,128 @@
...
+However, a package is free to define its main methods as the dispatchable
+versions, thereby allowing dispatch on the default implementation.
+
+Extensibility *and* Choice
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If some effort is put into the dispatch machinery, it's possible to
+dispatch over arbitrary objects --- including arrays, dtypes, and
+so on. A method defines the type of each dispatchable argument, and
+backends are *only* passed types they know how to dispatch over.
```
rgommers: This brings up some questions for me:

- How do you deal with functions that allow flexible inputs like `array_like`? 
- Does this lead to the need for using templating to cover a decent range of types? And does it suffer from combinatorial explosion, like when using Tempita + Cython?

docs/index.rst
```diff
@@ -3,8 +3,128 @@
...
+
+A user would simply do the following:
+
+.. code:: python
+
+    x = da.array([a, b, c])
+
+    # Anything that uses x will (usually) use the Dask backend
+
+    with ua.use_backend(DaskBackend):
```
rgommers: The combination of these examples makes me wonder what happen if a library skips a backend that the user forces. Does it fall back to a default. Does there *have to be* a default?

docs/index.rst
```diff
@@ -3,8 +3,128 @@
...
+        # Write all your code here
+        # It will prefer the Dask backend
+
+Addressing past flaws
+^^^^^^^^^^^^^^^^^^^^^
+
+The progress on NumPy's side for defining an override mechanism has been slow, with
+NEP-13 being first introduced in 2013, and with the wealth of dispatchable objects
+(including arrays, ufuns and dtypes), and with the advent of libraries like Dask,
+CuPy, XArray, PyData/Sparse and Xnd, it has become clear that the need for alternative
```
rgommers: Name capitalizations: Xarray, XND

docs/index.rst
```diff
@@ -3,8 +3,128 @@
...
+array-like implementations is growing. There are even other libraries like PyTorch, and
+TensorFlow that'd be possible to express in NumPy API-like terms. Another example includes
+the Keras API.
```
rgommers: The PyTorch and TensorFlow ones I get, although it should say "mostly". Keras is very different, so I would not use that here (functionality like `model.add/compile/evaluate/predict`, its generator classes, etc. don't have a NumPy equivalent).

docs/index.rst
hameerabbasi: So, my view is that `array_like` tries to do too much. What I've done is register just `ndarray` and allow coercion if someone really, really wants to use `array_like`.

hameerabbasi: I've also registered common things like `tuple, list, int, float, bool`, et. al. as "fallbacks", which means if they're encountered, NumPy will be tried again, but last.

docs/index.rst
hameerabbasi: There's no default. In the current code, it will be skipped. The skip is really just meant to be a way to avoid infinite recursion for libraries that both consume and export backends with intersecting method spaces.

docs/index.rst
hameerabbasi: What I mean is you could have a `ukeras` that depended on `unumpy`/`uscipy` which was also overridable. 🙂

docs/index.rst
hameerabbasi: To be more explicit: To `uarray`, NumPy is just a backend. There's no default at all, it'll just error if you don't select one. It's easy to add the capability for a default. You can provide a default implementation for methods (perhaps even in terms of others), but it can be overridden.

docs/index.rst
hameerabbasi: Ah, this part: It means whether it'll be asked to decide based on arrays, dtypes or ufuncs. I've made that more clear.

uarray/backend.py
```diff
@@ -9,11 +9,62 @@
...
+        This takes in args, kwargs and dispatchable args, and replaces all dispatchable arguments within
+        the args and kwargs, and then returns them.
+    default : Optional[Callable]
+        This is the default implementation for this multimethod, possibly in terms of other multimethods.
+
+    Examples
+    --------
+    >>> import uarray as ua
+    >>> def potato(a, b):
+    ...     return (a,) # b is not is dispatchable, so we return a only
```
rgommers: typo: remove second "is"

uarray/backend.py
```diff
@@ -9,11 +9,62 @@
...
+
+    Parameters
+    ----------
+    argument_extractor : ArgumentExtractorType
+        This is a callable that extracts the arguments using which dispatch can be performed from this
+        multimethod.
+    argument_replacer : ArgumentReplacerType
+        This takes in args, kwargs and dispatchable args, and replaces all dispatchable arguments within
+        the args and kwargs, and then returns them.
+    default : Optional[Callable]
```
rgommers: minor: the standard way to write this is `callable, optional`.

uarray/backend.py
```diff
@@ -9,11 +9,62 @@
...
+    """
+    This class is meant to provide a dispatchable method with or without a registered implementation.
+
+    Provides a multi-method instance. The method has no implementation by default, but one can be provided
+    if needed. :obj:`Backend` s can provide implementations for the method, and users can choose backends.
+
+    Parameters
+    ----------
+    argument_extractor : ArgumentExtractorType
+        This is a callable that extracts the arguments using which dispatch can be performed from this
```
rgommers: I had to read this more than once, so propose a small tweak: "... extracts the arguments that will be used to dispatch over". 

It would also help to state somewhere what happens if this returns more than one parameter. Do they all need to match? E.g. for the `potato` example, if `a` and `b` are both array kinds, does it ever make sense to use `b` (I think not)? And if `b` is a dtype for example, it would only dispatch to registered methods where both array and dtype match, and raise an error otherwise?

uarray/backend.py
```diff
@@ -9,11 +9,62 @@
...
+    >>> def potato_rd(args, kwargs, dispatch_args):
+    ...     # This replaces a within the args/kwargs
+    ...     return dispatch_args + args[1:], kwargs
+    >>> def potato_impl(a, b):
+    ...     # The default implementation passes through all arguments
+    ...     return a, b
+    >>> potato_mm = ua.MultiMethod(potato, potato_rd, default=potato_impl)
+    >>> potato_mm(1, '2')
+    (1, '2')
+    >>> be = ua.TypeCheckBackend((int,)) # Register implementation, define "supported dispatchable types"
```
rgommers: A little more text would help me here, something like: Register a second implementation, for which the supported type of `a` (I think, this is unclear - the rest of the example makes it look like it type checks `b`?) is only Python `int`.

uarray/backend.py
```diff
@@ -134,33 +179,188 @@ def inner(dispatcher: ArgumentExtractorType) -> MultiMethod:
...
 InstanceLookupType = Dict["DispatchableInstance", InstanceStubType]
 
 
 class Backend(metaclass=ABCMeta):
+    """
+    An abstract base class for all backend types.
+
+    See Also
+    --------
+    TypeCheckBackend: A specific kind of Backend.
```
rgommers: formatting: needs a space before `:` 

(is the case in more places, I'm not 100% sure if numpydoc/napoleon render it correctly otherwise)

uarray/backend.py
rgommers: Hmm, I'm confused now. Further down there's `Optional[bool], optional`. If `Optional` is a class then I don't see it so quickly.

uarray/backend.py
```diff
@@ -211,40 +426,75 @@ def usable(self, dispatchable_args: Tuple["DispatchableInstance"]) -> Optional[b
...
-set_backend = _SetBackend
-skip_backend = _SkipBackend
-
-
-def multimethod(backend: Backend, method: MultiMethod):
+def register_implementation(backend: Backend, method: MultiMethod):
+    """
+    Create an implementation for a given backend/method. The implementation
+    should have the same signature as the method, or perhaps with some optional
+    keyword arguments missing.
```
rgommers: only keyword arguments, so `f(a, b, *args)` is not supported? otherwise, drop "keyword"

uarray/backend.py
hameerabbasi: @rgommers `typing.Optional[T]` basically means that you can have `None` in addition to the allowed values of `T`. The `, optional` means that it's an optional argument in itself.

larsoner:
Azure says I am not authorized to view the documentation artifacts. Can you check on the permissions / make the artifacts public?

Another option would be to build docs with circle CI, which makes artifacts available with any GitHub login

rgommers:
> Azure says I am not authorized to view the documentation artifacts. Can you check on the permissions / make the artifacts public?

I can't either, I'm not sure that was intended in the first place. I think nothing is hooked up yet, but maybe readthedocs will be hooked up?

hameerabbasi:
> Azure says I am not authorized to view the documentation artifacts. Can you check on the permissions / make the artifacts public?

I'd like to, but I can't find way to make Azure artifacts public, from what I can tell. Neither documentation. It was intended, for easy review of documentation without having to build it yourself. 🙂

uarray/backend.py
```diff
@@ -25,44 +24,49 @@ class MultiMethod:
...
     --------
     >>> import uarray as ua
-    >>> def potato(a, b):
-    ...     return (a,) # b is not is dispatchable, so we return a only
-    >>> def potato_rd(args, kwargs, dispatch_args):
+
+    We first define an argument extractor, which extracts the relevant dispatchable arguments from the
+    arguments of the function. Note that the extracted arguments don't have to be direct arguments to
+    the function, they can be anything contained within as well. For example, if `a` was a `list` and
+    we wanted to treat everything within it as dispatchable, we could return `tuple(a)`.
```
rgommers: formatting: all these should be double backticks to render as code

hameerabbasi:
@larsoner You can view built docs at: https://uarray.readthedocs.io/en/latest/

You can also build them yourself with:

```bash
pip install -e .[docs]
sphinx-build -W -b html docs/ _build/html
```

And find the built docs in `./_build/html`.

PR #126, opened 2019-04-12
Minor fix to make pypi release render as markdown

costrouc:
Make eventual release of uarray on pypi render as markdown https://pypi.org/project/uarray/


hameerabbasi:
Thanks, @costrouc!

PR #135, opened 2019-04-17
Add docs for unumpy.

hameerabbasi:
@rgommers One minor comment, would you know how to make some text `monospace` in RST while also preserving the link?

### DIFF
--- a/docs/unumpy_using.rst
+++ b/docs/unumpy_using.rst
@@ -4,12 +4,25 @@ Using ``unumpy``
 ================
 
 ``unumpy`` is an in-progress mirror of the NumPy API which allows the user
-to dynamically switch out the backend that they are using. It also allows
+to dynamically switch out the backend that is used. It also allows
 auto-selection of the backend based on the arguments passed into a function.
+
+Note that currently, coverage is very incomplete. However, we have attempted
+to provide at least one of each kind of object in ``unumpy`` for
+reference. There are :obj:`ufunc` s and :obj:`ndarray` s,  which are classes,
+methods on :obj:`ufunc` such as , and also functions such as :obj:`sum` .
+
+The idea is that once things are more mature, it will be possible to switch
+out your backend with a simple import statement switch:
+
+.. code:: python
+
+    import numpy as np  # Old method
+    import unumpy as np  # Once this project is mature
 
 Currently, only the following functions are supported:
 
-* NumPy `universal functions <https://www.numpy.org/devdocs/reference/ufuncs.html>`_.
+* All NumPy `universal functions <https://www.numpy.org/devdocs/reference/ufuncs.html>`_.
 
   * `ufunc reductions <https://www.numpy.org/devdocs/reference/generated/numpy.ufunc.reduce.html#numpy.ufunc.reduce>`_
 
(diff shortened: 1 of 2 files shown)


rgommers:
> would you know how to make some text monospace in RST while also preserving the link?

I strongly suspect that that isn't possible without resorting to inserting raw html with `.. raw::`

docs/unumpy_using.rst
```diff
@@ -0,0 +1,35 @@
...
+Using ``unumpy``
+================
+
+``unumpy`` is an in-progress mirror of the NumPy API which allows the user
+to dynamically switch out the backend that they are using. It also allows
+auto-selection of the backend based on the arguments passed into a function.
+
+Currently, only the following functions are supported:
+
+* NumPy `universal functions <https://www.numpy.org/devdocs/reference/ufuncs.html>`_.
```
rgommers: I think this needs a bit more text. I actually read this as "all ufuncs are supported", but after looking at the implementation I think you just mean the `numpy.ufunc` class.

docs/unumpy_using.rst
```diff
@@ -0,0 +1,35 @@
+.. currentmodule:: unumpy
+
+Using ``unumpy``
+================
+
+``unumpy`` is an in-progress mirror of the NumPy API which allows the user
+to dynamically switch out the backend that they are using. It also allows
+auto-selection of the backend based on the arguments passed into a function.
+
```
rgommers: Before listing supported functions/objects, perhaps a bit more context here. Something like: "note: currently coverage is very incomplete, we are focusing on supporting one or a few of each *kind* of object - regular functions, ufuncs, classes and methods like `np.ufunc` and `np.ufunc.reduce`, `dtypes`, etc.".

docs/unumpy_using.rst
```diff
@@ -0,0 +1,35 @@
...
+* ufunc-based reductions
+
+  * `sum <https://docs.scipy.org/doc/numpy/reference/generated/numpy.sum.html>`_
+  * `prod <https://docs.scipy.org/doc/numpy/reference/generated/numpy.prod.html>`_
+  * `min <https://docs.scipy.org/doc/numpy/reference/generated/numpy.amin.html>`_
+  * `max <https://docs.scipy.org/doc/numpy/reference/generated/numpy.amax.html>`_
+  * `any <https://docs.scipy.org/doc/numpy/reference/generated/numpy.any.html>`_
+  * `all <https://docs.scipy.org/doc/numpy/reference/generated/numpy.all.html>`_
+
+The following backends are supported:
```
rgommers: Can you comment here on whether all listed objects above are supported for all backends? (for example I'm not sure if PyTorch has an exact `ufunc` equivalent).

docs/unumpy_using.rst
hameerabbasi: All `ufunc`s I found in the NumPy docs ARE supported. 🙂

rgommers: Ah, then I need to read the code better. Can you add "All" at the start of this line?

docs/unumpy_using.rst
```diff
@@ -4,12 +4,25 @@ Using ``unumpy``
...
 
 ``unumpy`` is an in-progress mirror of the NumPy API which allows the user
-to dynamically switch out the backend that they are using. It also allows
+to dynamically switch out the backend that is used. It also allows
 auto-selection of the backend based on the arguments passed into a function.
 
+Note that currently, coverage is very incomplete. However, we have attempted
+to provide at least one of each kind of object in ``unumpy`` for
+reference. There are :obj:`ufunc` s and :obj:`ndarray` s,  which are classes,
+methods on :obj:`ufunc` such as , and also functions such as :obj:`sum` .
```
rgommers: missing word in `such as ,`

docs/unumpy_using.rst
```diff
@@ -22,14 +35,14 @@ Currently, only the following functions are supported:
   * `any <https://docs.scipy.org/doc/numpy/reference/generated/numpy.any.html>`_
   * `all <https://docs.scipy.org/doc/numpy/reference/generated/numpy.all.html>`_
 
-The following backends are supported:
+You can use the :obj:`uarray.set_backend` decorator to set a backend and use the
+desired backend. Note that not every backend supports every method. For example,
+PyTorch does not have an exact :obj:`ufunc` equivalent, so we dispatch. The following
```
rgommers: "so we dispatch." --> I assume that needs to be "so we can't dispatch."

PR #142, opened 2019-05-07
add initial implementation for dask backend

MridulS:


rgommers:
Dask backend, awesome! Thanks @MridulS

PR #157, opened 2019-06-05
Remove canonicalization.

hameerabbasi:
Remove canonicalization (for now). Will come up with alternative strategies later. :slightly_smiling_face: 

Benchmarks: 30.7 microseconds (old) versus 2.06 microseconds (new).


rgommers:
15x faster, not bad:)

PR #160, opened 2019-06-11
WIP: Add documentation better separating out the developer segments

### LINKED ISSUE #159: Document the uarray API
After seeing on the scipy-dev mailing list that the API had changed significantly (now based on a protocol), I tried to understand the new API from https://uarray.readthedocs.io/en/latest/generated/uarray.html. That didn't help me. It needs a rewrite.

It should probably start with: what API does it allow offering to the **end user**?
- `set_backend` or `use_backend` (unclear which one, or both, right now)
- `set_global_backend` (is this for a user?)
- `from .... import libname_backend`
- anything else? (`BackendNotImplementedError`, other warnings/errors, does an end user import `uarray` or from somewhere else, etc.)

Then, what is the API for a **backend author**:
- function to register a backend
- `skip_backend`
- `__ua_domain__` and `__ua_function__` (is this for backend authors too?)
- `__ua_convert__` ?
- ...?

Finally, what is the API for a **library author** wanting to implement a backend system for her API?:
- `create_multimethod` / `generate_multimethod` (are both needed? if so, what is the difference?)
- `Dispatchable`
- `mark_as` (what's this?)
- ....

Another section that seems needed is _Concepts_. What are the fundamental concepts here?
- backend
- domain
- multimethod
- protocol
- dispatching
- coercion
- stacklevel (not in docs yet, but needed to explain how to get sane warnings)

### LINKED ISSUE #163: Add back the ability to register a backend
@peterbell10 As discussed in #159. Important questions:

1. Is both the ability to have a default backend necessary? 
2. Is having multiple registered backends necessary?
3. If only one is needed, which one?
4. Is the order important?

### LINKED ISSUE #166: Non-coercible Dispatchables
An observation based on the `unumpy` code
https://github.com/Quansight-Labs/uarray/blob/1fc6b68cbfbac12b6ff041a6a1a584feeaa8d2a8/unumpy/multimethods.py#L335-L338

If the `out` parameter is coerced to another type, then the result is not going to be written into the `out` param at all.

I think that either: `out` shouldn't really be a `Dispatchable`, or there needs to be some way of marking the parameter as non-coercible.

hameerabbasi:
Fixes #159
Fixes #166
Fixes #163

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -192,3 +192,7 @@ intersphinx_mapping = {
     "numpy": ("https://docs.scipy.org/doc/numpy/", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/reference/", None),
 }
+
+doctest_global_setup = """
+import uarray as ua
+"""
--- a/docs/enduser_docs.rst
+++ b/docs/enduser_docs.rst
@@ -42,10 +42,15 @@ It is a recommendation that the global backend should not depend on any
 other backend, as it is not guaranteed that another backend will be
 available.
 
+You can also register backends other than the global backend for permanent
+use, but the global backend will be tried first outside of a :obj:`set_backend`
+context. This can be done via :obj:`register_backend`.
+
+
 .. code:: python3
 
     import uarray as ua
 
-    ua.set_global_backend('mydomain', mybackend)
+    ua.set_global_backend(mybackend)
 
     # Use relevant multimethods here.
--- a/docs/enduser_docs.rst
+++ b/docs/enduser_docs.rst
@@ -1,0 +1,44 @@
+.. currentmodule:: uarray
+
+.. _userdocs:
+
+End-user quickstart
+===================
+
+Ideally, the only thing an end-user should have to do is set the backend and
+and its options. Given a backend, you (as the end-user) can decide to
+do one of two things:
+
+* Set the backend permanently (use the :obj:`set_global_backend` function).
+* Set the backend temporarily (use the :obj:`set_backend` context manager).
+
+Also of note may be the :obj:`BackendNotImplementedError`, which is raised
+when none of the selected backends have an implementation for a multimethod.
+
+Setting the backend temporarily
+-------------------------------
+
+To set the backend temporarily, use the :obj:`set_backend` context manager.
+
+.. code:: python3
+
+    import uarray as ua
+
+    with ua.set_backend(mybackend):
+        # Use multimethods (or code dependent on them) here.
+
+Setting the backend permanently
+-------------------------------
+
+To set the backend permanently, use the :obj:`set_global_backend`
+method. For this, you have to provide an additional piece of information,
+which is the domain. The domain, by convention, is the module where the
+multimethods reside, or one of its parents.
+
+.. code:: python3
+
+    import uarray as ua
+
+    ua.set_global_backend('mydomain', mybackend)
+
+    # Use relevant multimethods here.
--- a/docs/enduser_docs.rst
+++ b/docs/enduser_docs.rst
@@ -12,6 +12,9 @@ do one of two things:
 * Set the backend permanently (use the :obj:`set_global_backend` function).
 * Set the backend temporarily (use the :obj:`set_backend` context manager).
 
+.. note::
+    API authors may want to wrap these methods and provide their own methods.
+
 Also of note may be the :obj:`BackendNotImplementedError`, which is raised
 when none of the selected backends have an implementation for a multimethod.
 
@@ -34,6 +37,10 @@ To set the backend permanently, use the :obj:`set_global_backend`
 method. For this, you have to provide an additional piece of information,
 which is the domain. The domain, by convention, is the module where the
 multimethods reside, or one of its parents.
+
+It is a recommendation that the global backend should not depend on any
+other backend, as it is not guaranteed that another backend will be
+available.
 
 .. code:: python3
 
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -1,0 +1,50 @@
+.. currentmodule:: uarray
+
+Glossary
+========
+
+Multimethod
+-----------
+
+A method, possibly with a default/reference implementation, that can
+have other implementations provided by different backends.
+
+If a multimethod does not have an implementation, a
+:obj:`BackendNotImplementedError` is raised.
+
+Backend
+-------
+
+A backend is an entity that can provide implementations for different
+functions. It can also (optionally) receive some options from the user
+about how to process the implementations. A backend can be set permanently
+or temporarily.
+
+.. _DomainGlossary:
+
+Domain
+------
+
+A domain is a collection or grouping of multimethods. A domain's string,
+by convention (although not by force) is the name of the module that provides
+the multimethods.
+
+Dispatching
+-----------
+
+Dispatching is the process of choosing an implementation for a given multimethod.
+
+Coercion
+--------
+
+A backend might have different object types compared to the reference implementation,
+or it might require some other conversions of objects. Coercion is the process of doing
+forced conversions, ones that may take a long time.
+
+.. _MarkingGlossary:
+
+Marking
+-------
+
+Marking is the process of telling the backend what convertor to use for
+a given argument.
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -32,14 +32,24 @@ the multimethods.
 Dispatching
 -----------
 
-Dispatching is the process of choosing an implementation for a given multimethod.
+Dispatching is the process of forward a function call to an implementation
+in a backend.
+
+Conversion
+----------
+
+A backend might have different object types compared to the reference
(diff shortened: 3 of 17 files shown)


docs/conf.py
```diff
@@ -192,3 +192,7 @@
     "numpy": ("https://docs.scipy.org/doc/numpy/", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/reference/", None),
 }
+
+doctest_global_setup = """
+import uarray as ua
```
rgommers: comment mostly unrelated to this PR: I think it's actually better to just say `import uarray`. Library devs won't use uarray all the time, so explicit is better than saving 3 chars.

docs/enduser_docs.rst
```diff
@@ -0,0 +1,44 @@
...
+To set the backend permanently, use the :obj:`set_global_backend`
+method. For this, you have to provide an additional piece of information,
+which is the domain. The domain, by convention, is the module where the
+multimethods reside, or one of its parents.
+
+.. code:: python3
+
+    import uarray as ua
+
+    ua.set_global_backend('mydomain', mybackend)
```
rgommers: I guess a library would expose this instead normally? E.g. this would become 
```
from scipy import fft

fft.set_global_backend(somebackend)
```
?

docs/glossary.rst
```diff
@@ -0,0 +1,50 @@
...
+------
+
+A domain is a collection or grouping of multimethods. A domain's string,
+by convention (although not by force) is the name of the module that provides
+the multimethods.
+
+Dispatching
+-----------
+
+Dispatching is the process of choosing an implementation for a given multimethod.
```
rgommers: rather than "choosing", perhaps "forward a function call to an implementation in a backend"?

docs/glossary.rst
```diff
@@ -0,0 +1,50 @@
...
+or it might require some other conversions of objects. Coercion is the process of doing
+forced conversions, ones that may take a long time.
+
+.. _MarkingGlossary:
+
+Marking
+-------
+
+Marking is the process of telling the backend what convertor to use for
+a given argument.
```
rgommers: add "when coercing"?

docs/libauthor_docs.rst
```diff
@@ -0,0 +1,60 @@
+.. currentmodule:: uarray
+
+.. _libauthordocs:
+
+Documentation for backend providers
+===================================
+
+Backend providers need to be aware of three protocols: ``__ua_domain__``,
+``__ua_function__`` and ``__ua_convert__``. These three must be a part of
+every backend object.
```
rgommers: the docs are clear, it just raises the question why? I imagine that coercion would not make sense for many objects.

docs/libauthor_docs.rst
```diff
@@ -0,0 +1,60 @@
...
+
+``__ua_domain__`` is a string containing the domain of the backend. This is,
+by convention, the name of the module (or one of its dependencies or parents)
+that contain the multimethods. For example, ``scipy`` and ``numpy.fft`` could
+both use the ``numpy`` domain.
+
+``__ua_function__``
+-------------------
+
+This is the most important protcol, one that defines the implementation of a
```
rgommers: typo: `protcol` (and same typo also further down)

docs/multimethod_docs.rst
```diff
@@ -0,0 +1,71 @@
...
+signature as the original multimethod. It can also be used to provide one
+multimethod in terms of others, even if the default implementation for the.
+downstream multimethods is not defined.
+
+Examples
+--------
+
+Examples of writing multimethods are found in `this file <https://github.com/Quansight-Labs/uarray/blob/master/unumpy/multimethods.py>`_.
+It also teaches some advanced techniques, such as overriding instance methods,
+``__call__``. The same philosophy may be used to override properties, static
```
rgommers: this sentence is a bit odd. end with "instance methods, e.g., `__call__`"?

docs/multimethod_docs.rst
```diff
@@ -0,0 +1,71 @@
+.. currentmodule:: uarray
+
+.. _mmauthordocs:
+
+Documentation for multimethod authors
```
rgommers: This section is hard to understand. To start with the title: this is aimed at library authors wanting to provide a backend system, so other packages can implement a backend. That's not really conveyed by the title or the text below.

Then, it misses a description of what that backend will actually look like and how to implement it. Concretely, if say `scipy.fft` wants to implement a backend system, how does it provide its context manager and multimethods? That's the most important part. Once you have that, then go into how to write a single multimethod.

docs/conf.py
hameerabbasi: The protocols are also all `__ua_*__`... And changing those would mean changing the API (again). I'm doing this for uniformity...

docs/enduser_docs.rst
hameerabbasi: No, the library would have to do something like `def set_fft_backend(be): ua.set_global_backend('scipy.fft', be)`

hameerabbasi: I can write docs on how to do that though.

docs/libauthor_docs.rst
hameerabbasi: Ah, yes... So the reverse dispatcher and the coercion wouldn't even be necessary. 😄

docs/libauthor_docs.rst
```diff
@@ -0,0 +1,65 @@
...
+``__ua_domain__`` is a string containing the domain of the backend. This is,
+by convention, the name of the module (or one of its dependencies or parents)
+that contain the multimethods. For example, ``scipy`` and ``numpy.fft`` could
+both use the ``numpy`` domain.
+
+``__ua_function__``
+-------------------
+
+This is the most important protocol, one that defines the implementation of a
+multimethod. It has the signature ``(method, args, kwargs, dispatchable_args)``.
```
peterbell10: It's not clear to me what a backend is meant to do with `dispatchable_args`. Isn't the job of `argument_replacer` to rewrite the function call arguments such that this doesn't matter?

hameerabbasi: When `__ua_convert__` is not present, the purpose is to recognize whether these arguments can be handled or not.

peterbell10: Is it not enough for them to be in `args` and `kwargs`?

hameerabbasi: You don't know which ones in `args`/`kwargs` are dispatchable and which ones are standard arguments, for a generalized function call. I have added documentation to that effect. I also made `__ua_convert__` mandatory.

docs/glossary.rst
hameerabbasi: I've clarified the difference above.

docs/libauthor_docs.rst
hameerabbasi: If you prefer this functionality be moved into `__ua_convert__`, that'd be fine by me too.

peterbell10: I think it's better to have a clear seperation of responsibilities here.

docs/libauthor_docs.rst
```diff
@@ -0,0 +1,70 @@
...
+``(value, dispatch_type, coerce)``. ``value`` is the value to convert,
+``dispatch_type`` is the mark of the object to be converted, and ``coerce``
+specifies whether or not to "force" the conversion. By convention, operations
+larger than ``O(log n)`` (where ``n`` is the size of the object in memory)
+should only be done if ``coerce`` is ``True``.
+
+If ``__ua_convert__`` returns ``NotImplemented``, it means that the backend in
+question does not support converting the given ``dispatch_type``, which signals
+that the value should be passed on as-is and treated as though it wasn't
+dispatchable.
```
peterbell10: I still think that there needs to be a way to say "the backend doesn't support this argument". As an example: what should a CuPy backend return when passed a NumPy array and `coerce=False`?

peterbell10: ...of course, it could just return the NumPy array but then it would need to do redundant type-checking in its `__ua_function__`.

hameerabbasi: How about a sentinel: `ua.TypeNotImplemented` for the current behaviour, `NotImplemented` for "I don't want to handle this, fail", and we remove `dispatchable_args` completely.

hameerabbasi: I believe that'd satisfy all our concerns. :slightly_smiling_face:

peterbell10: For your case of `ua.TypeNotImplemented`, why does this need to be a sentinel and not just have the `__ua_convert__` function pass the type through unchanged?

hameerabbasi: Ah, yes. My original purpose was the removal from `dispatchable_args`, but that's covered already.

hameerabbasi: This is all done, thanks for the awesome feedback!

docs/libauthor_docs.rst
hameerabbasi: Done!

PR #162, opened 2019-06-11
Fix normalisation.

### LINKED ISSUE #161: Is argument normalisation really needed?
Let's say we have a Python function

```python
def transform_args_kwargs(f):
    @functools.wraps(f)
    def wrapped(a, b=None):
        # Do something with args/kwargs
        # Discard default arguments
        # return args/kwargs -- Note that kwargs will contain `b`.
    return wrapped

@transform_args_kwargs
def f(a, b=None):
     return a, b
```

Basically, #157 locks us into using the multimethod only in the form `f(something, b=something)`, but not any other form.

hameerabbasi:
Fixes #161

### DIFF
--- a/uarray/backend.py
+++ b/uarray/backend.py
@@ -93,6 +93,7 @@ def generate_multimethod(
     uarray
         See the module documentation for how to override the method by creating backends.
     """
+    defaults, opts = get_defaults(argument_extractor)
 
     @functools.wraps(argument_extractor)
     def inner(*args, **kwargs):
@@ -164,6 +165,12 @@ def generate_multimethod(
                 )
 
         args, kwargs = argument_replacer(args, kwargs, tuple(replaced_args))
+
+        kwargs = {k: kwargs[k] for k in opts if k in kwargs}
+        for k, v in defaults.items():
+            if k in kwargs and kwargs[k] is v:
+                del kwargs[k]
+
         return args, kwargs, filtered_args
 
     inner._coerce_args = replace_dispatchables  # type: ignore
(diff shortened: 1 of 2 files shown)


uarray/backend.py
```diff
@@ -164,6 +165,12 @@ def replace_dispatchables(
                 )
 
         args, kwargs = argument_replacer(args, kwargs, tuple(replaced_args))
+
+        kwargs = {k: kwargs[k] for k in opts if k in kwargs}
```
peterbell10: This doesn't work because opts contains `Parameter` objects and compares them to strings. Instead, I think `opts` should be `defaults.keys()`.

That being said, is it even necessary? Shouldn't the call to `argument_extractor` fail if there are invalid `kwargs`. At least it did in my case.

hameerabbasi:
Can you post some test code which takes that long? I'll benchmark it. Are you measuring the performance of just the line calling the multimethod or also setting the backend? Setting it takes a bit but calling it has very low overhead.

uarray/backend.py
hameerabbasi: `opts` was meant to include the list of optional arguments, so that argument replacers could be unified... For example, `all` has no `dtype` arg, but `sum` does. Otherwise it'd require separate argument replacers for both.

If you prefer the other method, I'm willing to revert this change.

peterbell10:
My original benchmark was measuring setup once and calling many thousands of times. In the interest of fairness, I've changed the benchmark to only measure calls and to do better statistics. I'm now measuring overheads of ~9 us per function call.
The following benchmark was run against my `scipy.fft` uarray code (peterbell10/scipy#2):
```python
from timeit import timeit
import numpy as np

def time_stats(stmt, setup, number):
    N = 20
    ts = np.zeros(N)
    for i in range(N):
        ts[i] = timeit(setup=setup, stmt=stmt, number=number//N)
    return np.sum(ts), np.std(ts)*np.sqrt(N)


setup = '''
import scipy.fft
import numpy as np

x = np.arange(10)
'''
n=1000000
print('Indirect:', time_stats('scipy.fft.fft(x)', setup, n))
print('Direct  :', time_stats('scipy.fft._pocketfft.fft(x)', setup, n))
```

uarray/backend.py
hameerabbasi: I've fixed this in fd30f61144a7149b844e6b617941873df3f40abe.

hameerabbasi:
Can you possibly test without this PR in place? I doubt it'll make much of a difference, as the `inspect` module is used only once.

peterbell10:
Okay, I tested both with and without this PR and they were about the same overhead.

hameerabbasi:
In that case, I'll go ahead and merge this. However, I'll still try to bring the overhead down.

Issue #164, opened 2019-06-18
Proposal: New argument passing convention


peterbell10:
## Adding new arguments to functions
**Problem**:
Currently, `__ua_function__` takes its arguments by the standard python `args, kwargs` pair. However, I think this creates a problem under maintenance. To illustrate this, consider that our backend supports a simple `fft` function
```python
def fft(x)
```
and in a new version of scipy, we want to add support for different normalization modes via a `norm` argument
```python
def fft(x, norm=None)
```

Because the argument has been added in a backwards compatible way, it's still fine to call `fft(x)` on the old backend which doesn't support the `norm` argument. Because of #162, it's also fine to explicitly call `fft(x, norm=None)`. However `fft(x, None)` doesn't work because positional arguments are not canonicalized.

**Solution**: convert all arguments into one dict and canonicalize all of them.

## Adding hint arguments
As an extension, we can also consider another kind of argument
```python
def fft(x, norm=None, overwrite_x=False)
```
in `scipy.fft`, `overwrite_x` is a hint to the implementation that it can reuse the memory from the input array `x`. Importantly, there's no obligation for the implementation to actually respect the argument and it doesn't effect the result of the function.

In other words, its actually okay to use a backend that doesn't know about this argument, even if called with `fft(x, overwrite_x=True)`.

So, I propose spitting the argument dictionary into two dicts: `args` and `hints`. If the backend sees a key in `args` that it doesn't recognise, it should return `NotImplemented`, but any unknown `hints` can be safely ignored.

hameerabbasi:
> Solution: convert all arguments into one dict and canonicalize all of them.

While this is the common case, there are positional-only and keyword-only arguments... So keeping `args` around is needed. One dict won't work, but we can move every _possible_ arg into the `dict`.

peterbell10:
Positional only arguments still have names so I don't see why they couldn't be inserted into a `dict`.

hameerabbasi:
They do... But they couldn't be called in the form `f(**kwargs)`.

peterbell10:
Positional-only arguments only matter for the _caller-facing_ signature (`argument_extractor`) so there's no reason why the _backend_ couldn't be callable as `f(**kwargs)`.

hameerabbasi:
> Positional-only arguments only matter for the _caller-facing_ signature (`argument_extractor`) so there's no reason why the _backend_ couldn't be callable as `f(**kwargs)`.

[...]

PR #167, opened 2019-06-19
ENH: Pass all arguments as kwargs

peterbell10:
Implements the first part of #164 

All positional arguments are converted into a dictionary and used to update `kwargs`. varargs are stored in the `kwargs` dictionary as a tuple e.g.

```python
def min(a, *args):
	pass
```
gives kwargs of the form
```python
{'a', a, 'args': (arg1, arg2, ...)}
```

### DIFF
--- a/docs/libauthor_docs.rst
+++ b/docs/libauthor_docs.rst
@@ -26,12 +26,13 @@ both use the ``numpy`` domain.
 -------------------
 
 This is the most important protocol, one that defines the implementation of a
-multimethod. It has the signature ``(method, args, kwargs)``.
-Note that it is called in this form, so if your backend is an object instead of
-a module, you should add ``self``. ``method`` is the multimethod being called,
-and it is guaranteed that it is in the same domain as the backend. ``args`` and
-``kwargs`` are the arguments to the function, possibly after conversion
-(explained below)
+multimethod. It has the signature ``(method, kwargs)``. Note that it is called
+in this form, so if your backend is an object instead of a module, you should
+add ``self``. ``method`` is the multimethod being called, and it is guaranteed
+that it is in the same domain as the backend. ``kwargs`` is a ``dict``
+containing all arguments to the function, possibly after conversion (explained
+below). If the function contains any varargs (e.g. ``def f(*args)``) then it is
+given in ``kwargs`` as a single parameter.
 
 Returning :obj:`NotImplemented` signals that the backend does not support this
 operation.
--- a/unumpy/multimethods.py
+++ b/unumpy/multimethods.py
@@ -407,15 +407,15 @@ def var(a, axis=None, dtype=None, out=None, ddof=0, keepdims=False):
 
 
 # set routines
-@create_numpy(_generic_argreplacer("a"))
+@create_numpy(_generic_argreplacer("ar"))
 @all_of_type(ndarray)
-def unique(a, return_index=False, return_inverse=False, return_counts=False, axis=None):
-    return (a,)
+def unique(ar, return_index=False, return_inverse=False, return_counts=False, axis=None):
+    return (ar,)
 
 
-@create_numpy(_generic_argreplacer("element", "test_elements"))
+@create_numpy(_generic_argreplacer("ar1", "ar2"))
 @all_of_type(ndarray)
-def in1d(element, test_elements, assume_unique=False, invert=False):
+def in1d(ar1, ar2, assume_unique=False, invert=False):
     return (element, test_elements)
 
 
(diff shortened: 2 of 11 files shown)


docs/libauthor_docs.rst
```diff
@@ -26,12 +26,13 @@ both use the ``numpy`` domain.
...
-Note that it is called in this form, so if your backend is an object instead of
-a module, you should add ``self``. ``method`` is the multimethod being called,
-and it is guaranteed that it is in the same domain as the backend. ``args`` and
-``kwargs`` are the arguments to the function, possibly after conversion
-(explained below)
+multimethod. It has the signature ``(method, kwargs)``. Note that it is called
+in this form, so if your backend is an object instead of a module, you should
+add ``self``. ``method`` is the multimethod being called, and it is guaranteed
+that it is in the same domain as the backend. ``kwargs`` is a ``dict``
+containging all arguments to the function, possibly after conversion (explained
```
hameerabbasi:
```suggestion
containing all arguments to the function, possibly after conversion (explained
```

unumpy/multimethods.py
```diff
@@ -506,56 +495,49 @@ def broadcast_arrays(*args, subok=False):
...
 
-def _first_argreplacer(args, kwargs, arrays1):
-    def func(arrays, axis=0, out=None):
-        return (arrays1,), dict(axis=0, out=None)
-
-    return func(*args, **kwargs)
-
-
-@create_numpy(_first_argreplacer)
+@create_numpy(_generic_argreplacer('arrays'))
```
hameerabbasi: See comment for `stack`.

unumpy/multimethods.py
```diff
@@ -506,56 +495,49 @@ def broadcast_arrays(*args, subok=False):
...
 @all_of_type(ndarray)
 def concatenate(arrays, axis=0, out=None):
     return arrays
 
 
-@create_numpy(_first_argreplacer)
+@create_numpy(_generic_argreplacer('arrays'))
```
hameerabbasi: Here, you're replacing the whole argument with one array. You must replace `arrays` with a list of all passed-in arrays.

unumpy/multimethods.py
```diff
@@ -497,7 +488,8 @@ def lexsort(keys, axis=None):
 
 
 def _args_argreplacer(args, kwargs, arrays):
```
hameerabbasi:
```suggestion
def _args_argreplacer(kwargs, arrays):
```

peterbell10:
>Looks like unfortunately NumPy _expects_ posargs, and so without a small fix to `__ua_function__` to move a or self to posargs, this won't work.

I don't follow, can you give an example?

hameerabbasi:
```python
   def __ua_function__(method, kwargs):
        args = ()
        if 'self' in kwargs:
            args = args + (kwargs.pop('self'),)
        if 'a' in kwargs:
            args = args + (kwargs.pop('a'),)
        if method in _implementations:
            return _implementations[method](*args, **kwargs)

        if not hasattr(cp, method.__name__):
            return NotImplemented

        return getattr(cp, method.__name__)(**kwargs)
```

peterbell10:
I think it should be work without that, it's just that the argument names have to match now.

hameerabbasi:
> I think it should be work without that, it's just that the argument names have to match now.

That's a fair point. 🙂

hameerabbasi:
Looks like Dask and Sparse still have issues. 🙍‍♂

hameerabbasi:
I wonder if all this trouble is worth it... It seemed like a good idea in theory. I would like to propose the following change:

* We keep around args, but normalize so that any default args are removed if they're at the end of `*args`.

peterbell10:
Yes, it looks like the different libraries use different names for some parameters. In python 3.8, these would probably be marked as [positional-only parameters][PEP].

How about adding a parameter to `generate_multimethod` that gives the number of positional-only parameters. That way positional-only parameters can still be defaulted.

[PEP]: https://www.python.org/dev/peps/pep-0570/

hameerabbasi:
I'm proposing something different, but a bit more magical: We store the defaults by position as well as by keyword. Then, we trim `*args` from the end.

peterbell10:
I understood, I am just suggesting an alternative :)

If the argument is conceptually positional-only then I think that the current backends will not work properly when they're passed as `kwargs` so it may be worth considering.

hameerabbasi:
> I understood, I am just suggesting an alternative :)

It's a valid alternative. When Python 3.8 comes along and is widely used, we can probably move to it... But in the meanwhile, adding an extra parameter adds overhead for API authors that they may not be willing to foot.

hameerabbasi:
Would you like to keep this open for the future?

peterbell10:
I'll just close for now.

PR #168, opened 2019-06-20
Canonicalize positional arguments

peterbell10:
As suggested in #167

### DIFF
--- a/uarray/backend.py
+++ b/uarray/backend.py
@@ -107,14 +107,14 @@ def generate_multimethod(
     uarray
         See the module documentation for how to override the method by creating backends.
     """
-    defaults, opts = get_defaults(argument_extractor)
+    kw_defaults, arg_defaults, opts = get_defaults(argument_extractor)
 
     @functools.wraps(argument_extractor)
     def inner(*args, **kwargs):
-        # args, kwargs = _canonicalize(argument_extractor, args, kwargs)
         dispatchable_args = argument_extractor(*args, **kwargs)
         errors = []
 
+        args = canonicalize_args(args)
         result = NotImplemented
 
         for options in _backend_order(domain):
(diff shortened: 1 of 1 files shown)


uarray/backend.py
```diff
@@ -174,12 +174,25 @@ def replace_dispatchables(
...
         kwargs = {k: kwargs[k] for k in opts if k in kwargs}
-        for k, v in defaults.items():
+        for k, v in kw_defaults.items():
             if k in kwargs and kwargs[k] is v:
                 del kwargs[k]
 
         return args, kwargs
 
+    def canonicalize_args(args):
+        if len(args) > len(arg_defaults):
```
hameerabbasi: Isn't this branch already handled by the `return` at the end of this function?

uarray/backend.py
```diff
@@ -174,12 +174,25 @@ def replace_dispatchables(
         args, kwargs = argument_replacer(args, kwargs, tuple(replaced_args))
 
         kwargs = {k: kwargs[k] for k in opts if k in kwargs}
```
hameerabbasi: It might be nice to factor this out into the same function as `*args`.

uarray/backend.py
```diff
@@ -309,14 +322,20 @@ def skip_backend(backend):
...
     sig = inspect.signature(f)
-    defaults = {}
+    kw_defaults = {}
+    arg_defaults = []
     opts = set()
     for k, v in sig.parameters.items():
         if v.default is not inspect.Parameter.empty:
-            defaults[k] = v.default
+            kw_defaults[k] = v.default
+        if v.kind in (
```
hameerabbasi: Nice check. ;) I wasn't sure you'd add it on the first pass.

uarray/backend.py
peterbell10: The issue is with this line
```python
for a, d in zip(args[::-1], arg_defaults[len(args) - 1 :: -1]):
```
the slice of `arg_defaults` will not match up with `args`.

uarray/backend.py
peterbell10: This step is done after argument replacement whereas I have put my step before `args` reaches the backend. This way it's only done once.

hameerabbasi:
Are you going to make any more changes? If not I'll merge. 😄

peterbell10:
I think it's good

hameerabbasi:
Thanks for this, @peterbell10!

PR #170, opened 2019-06-24
ENH: Implement multimethod call in the Python C API

### LINKED ISSUE #156: Measure uarray overhead and improve it
Uarray overhead is currently on the order of 100 us. This is too high - it will prevent adoption as the tool to define one's main API for many use cases.

Would be good to be able to measure the overhead easily:
- test setup
- benchmarks

And then bring it down to a more reasonable level. What is possible without removing functionality? What is the most expensive part of the current design?

peterbell10:
I've been working on a C++ implementation of the main uarray multimethod call. I'm not familliar with the C API but this works well enough for my scipy tests to pass.

In my benchmarks the overhead is down from ~5 us to ~2.5 us. Fixes #156.


setup.py
```diff
@@ -43,6 +43,11 @@ def parse_requires():
 with open("README.md") as f:
     long_desc = f.read()
 
+
+extensions = [
+    Extension('uarray._uarray', sources=['uarray/_uarray_dispatch.cxx'], language='c++')
```
hameerabbasi: This might fix it. 😄 

I also implemented your suggestions during the meeting in #169.

```suggestion
    Extension('uarray._uarray', sources=['uarray/_uarray_dispatch.cxx'], language='c++11')
```

setup.py
```diff
@@ -43,6 +43,11 @@ def parse_requires():
 with open("README.md") as f:
     long_desc = f.read()
 
+
+extensions = [
+    Extension('uarray._uarray', sources=['uarray/_uarray_dispatch.cxx'], language='c++11')
```
hameerabbasi: Interesting...

```python
```suggestion
    Extension('uarray._uarray', sources=['uarray/_uarray_dispatch.cxx'], language='c++11', extra_compile_args=['-std=c++11'])
```

```

**Comment by hameerabbasi** [comment]
Ah, I see you discovered the descriptor protocol. 😅 That's a tough one.

**Comment by peterbell10** [comment]
I spent way too much time trying to figure out why adding a method called `__get__` did nothing...

File: uarray/_uarray_dispatch.cxx [hunk 3]
```diff
@@ -0,0 +1,479 @@
...
+
+
+/** Implements the descriptor protocol to allow binding to class instances */
+PyObject * Function_descr_get(PyObject * self, PyObject * obj, PyObject * type)
+{
+	Py_INCREF(self);
+	if (!obj)
+		return self;
+
+	return PyMethod_New(self, obj);
```
hameerabbasi: I believe you have to use a new "object" that implements the descriptor protocol, by injecting the args/kwargs.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,479 @@
...
+
+	return PyUnicode_FromString("<uarray multimethod>");
+}
+
+
+/** Implements the descriptor protocol to allow binding to class instances */
+PyObject * Function_descr_get(PyObject * self, PyObject * obj, PyObject * type)
+{
+	Py_INCREF(self);
+	if (!obj)
```
hameerabbasi: You might want to compare to `Py_None` here.

peterbell10: If I call `func.__get__(None)` this receives `nullptr`.

uarray/_uarray_dispatch.cxx
peterbell10: What do you mean? This _does_ implement the descriptor protocol.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,479 @@
...
+	}
+
+	return {std::move(new_args), std::move(new_kwargs)};
+}
+
+
+PyObject * uarray_function_call(
+	PyObject * self, PyObject * args, PyObject * kwargs)
+{
+	return reinterpret_cast<Function *>(self)->call(args, kwargs);
```
hameerabbasi: Here, I believe you need to inject `obj` into the `args` if it isn't `nullptr`.

peterbell10: The `PyMethod` wrapper already takes care of that:
```
>>> class Foo: pass
>>> Foo.f = func
>>> Foo().f
<bound method extractor of <__main__.Foo object at 0x7f0f9d128630>>
```

peterbell10:
The weird thing is that I can reproduce the crash when doing a full pytest run, but none of the individual tests crash...  Posibly a use after free that's GC dependant somewhere?

hameerabbasi:
I usually use `valgrind` or a debugger like `lldb` to find the line where it crashes and work backwards from there.

hameerabbasi:
(Not the most fun thing in the world)

peterbell10:
I already know it's a segfault in numpy's `get_ufunc_arguments`.

hameerabbasi:
So we know one of two things: Either it's in NumPy, or it's the `ufunc` or the arguments. Since you're handling the args fine, you probably need something for the `ufunc`, or `self` in this case.

uarray/_uarray_dispatch.cxx
hameerabbasi: Probably need a `Py_INCREF` for `obj` here.

hameerabbasi:
> The weird thing is that I can reproduce the crash when doing a full pytest run, but none of the individual tests crash... Possibly a use after free that's GC dependant somewhere?

That is weird indeed... Either one is doing too many `incref`s, too many `decref`s. Python has a refcounting GC, I don't know how cycles are handled in such a GC, maybe a delayed pass?

hameerabbasi:
I guess we shouldn't be getting circular references at all, but here's a reference:https://stackoverflow.com/questions/10962393/how-does-pythons-garbage-collector-detect-circular-references

Not that it helps in a use-after-free.

peterbell10:
I've actually tried removing all of the `Py_DECREF`s from my code and it still segfaults. Maybe this isn't use after free at all.

peterbell10:
Unless I'm not `INCREF`ing something.

uarray/_uarray_dispatch.cxx
```diff
@@ -232,35 +233,26 @@ py_func_args Function::replace_dispatchables(
...
 	auto replacer_args = py_make_tuple(args, kwargs, replaced_args);
 	if (!replacer_args)
 		return {};
 
 	res = py_ref::steal(PyObject_Call(replacer_, replacer_args, nullptr));
-	if (PyTuple_Size(res) != 2)
+	if (!res)
+		return {};
+
+	if (!PyTuple_Check(res) || PyTuple_Size(res) != 2)
```
hameerabbasi: You might want to use `Py_BuildValue` and `PyArg_ParseTuple` more. Although it might bring up the overhead, not sure... https://docs.python.org/3/c-api/arg.html

hameerabbasi:
Maybe we're somehow... Using `steal` where there should be a `borrow`?

peterbell10:
I've tried changing all `steal`s to `ref`s and it still happens.

hameerabbasi:
Then let's try changing gears a bit... Is it in upstream NumPy? Are we supposed to INCREF before passing things into a `ufunc` somehow? Seems unlikely, since it's done from Python.

hameerabbasi:
Two more things: A `dict` returns NULL instead of setting an exception if it's accessed with an invalid key, so it might be a null pointer. And a tuple-access, out-of-bounds might do it.

Speaking of, the first could be the reason, given that it only happens on this repo where kwargs are actually being dropped.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,491 @@
...
+{
+	if (kwargs == nullptr)
+		return py_ref::steal(PyDict_New());
+
+	PyObject * key, * def_value;
+	Py_ssize_t pos = 0;
+	while (PyDict_Next(def_kwargs_, &pos, &key, &def_value))
+	{
+		auto val = PyDict_GetItem(kwargs, key);
+		if (is_default(val, def_value))
```
hameerabbasi: Add `val != nullptr` here?

peterbell10: It doesn't fix it and I don't think that should matter since `def_value` shoudn't be `nullptr`.

peterbell10:
For me, the tests always seem to fail at 
```
unumpy/tests/test_numpy.py::test_functions_coerce[backend1-method10-args10-kwargs10]
```

Could there be something special about `unumpy.var` or `DaskBackend`

uarray/_backend.py
```diff
@@ -109,93 +106,17 @@ def generate_multimethod(
...
-            return NotImplemented
-
-        return argument_replacer(args, kwargs, tuple(replaced_args))
-
-    def canonicalize_args(args, kwargs):
-        if len(args) > len(arg_defaults):
-            return args
-
-        match = 0
-        for a, d in zip(args[::-1], arg_defaults[len(args) - 1 :: -1]):
```
hameerabbasi: Look here for inspiration.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,491 @@
...
+	const auto def_size = PyTuple_GET_SIZE(def_args_.get());
+
+	if (arg_size > def_size)
+		return py_ref::ref(args);
+
+	Py_ssize_t mismatch = 0;
+	for (Py_ssize_t i = arg_size - 1; i >= 0; --i)
+	{
+		auto val = PyTuple_GET_ITEM(args, i);
+		auto def = PyTuple_GET_ITEM(def_args_.get(), i);
```
hameerabbasi: There might be a negative access here since `def_args_` isn't sliced properly.

peterbell10: The check above guaruntees that `def_size >= arg_size`

hameerabbasi: Hmm, it's a bug in any case, and should be fixed.

peterbell10: Not sure I follow, what's a bug?

hameerabbasi:
I ran this locally, and I couldn't reproduce the segfault. But I see a lot of `SystemError: <uarray multimethod 'something'> returned NULL without setting an error.` bugs.

uarray/_uarray_dispatch.cxx
hameerabbasi: Never mind, mind derp.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,491 @@
...
+		{
+			Py_XINCREF(object);
+			return py_ref(object);
+		}
+
+	~py_ref(){ Py_XDECREF(obj); }
+
+	py_ref & operator = (const py_ref & other)
+		{
+			py_ref(other).swap(*this);
```
hameerabbasi: Is this right? Shouldn't you just be copying and `INCREF`-ing `other.obj`?

uarray/_uarray_dispatch.cxx
hameerabbasi: Shouldn't this be offset by `len(args)`?

uarray/_uarray_dispatch.cxx
peterbell10: It's the [Copy-Swap idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-and-swap).
The temporary is copy constructed (which calls `Py_XINCREF`) and then the pointer is swapped into place. When the temporary is destroyed, the old value is `Py_XDECREF`ed.

hameerabbasi:
What system are you on? Maybe I can debug this... But it doesn’t segfault on my system.

peterbell10:
I'm running ubuntu 18.04 with python 3.7.1

peterbell10:
>I ran this locally, and I couldn't reproduce the segfault. But I see a lot of SystemError: <uarray multimethod 'something'> returned NULL without setting an error. bugs.

The `BackendNotImplementedError` variable was being reset in my init code. Should be fixed now.

azure-pipelines.yml
```diff
@@ -16,6 +16,7 @@ jobs:
 
   - script: |
       source activate uarray
+      python3 -X dev `which pytest` -v
       pytest
```
hameerabbasi: This line should be removed.

hameerabbasi:
I can't reproduce this on Linux in Conda or macOS in conda with the latest Python and NumPy.

peterbell10:
What version of dask are you using?

hameerabbasi:
2.0.

hameerabbasi:
Could you try removing the `xnd` backend?

peterbell10:
I have it removed locally and it still crashes. Although I just realised my earlier comment was a bit confusing
>unumpy/tests/test_numpy.py::test_functions_coerce[backend1-method10-args10-kwargs10]

here backend1 should be the dask backends since I've removed the xnd one.

peterbell10:
Sometimes I also see the segfault from the sparse backend, but it's also in an `__array_ufunc__` override.

hameerabbasi:
Remove everything BUT Dask?

hameerabbasi:
I wonder if this is Numba being weird with references...

peterbell10:
The stacktrace I get from dask looks like infinite recusion:
```
Fatal Python error: Segmentation fault
Current thread 0x00007f5ec7840740 (most recent call first):
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/core.py", line 300 in apply_infer_dtype
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/core.py", line 3663 in elemwise
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/core.py", line 1090 in __array_ufunc__
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/core.py", line 3743 in _enforce_dtype
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/utils.py", line 122 in compute_meta
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/blockwise.py", line 233 in blockwise
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/core.py", line 3684 in elemwise
File "/home/peter/.local/lib/python3.7/site-packages/dask/array/core.py", line 1090 in __array_ufunc__
[...]
```

hameerabbasi:
Ahh... A stack overflow. ☹️ 

Can we figure out the args we pass in and report upstream?

peterbell10:
Oh, I only just realised that [master is failing as well](https://dev.azure.com/Quansight-Labs/uarray/_build/results?buildId=149). Could it be from the changes in d179d70665ebdc4bf or maybe a dependency updated?

hameerabbasi:
Dask 2.0 is indeed the breaking update.

hameerabbasi:
This is weird: 

```python
>>> import dask
>>> dask.__version__
'2.0.0'
```

And PyTest succeeds.

hameerabbasi:
Is the breaking change in a sparse now?

peterbell10:
I've looked into it and I think sparse is encountering a bug in NumPy's fancy indexing.  I was able to reproduce the segfault on master and it only happens with `-X dev`. I think this is because python's debug allocator overwrites freed memory with `0xdb`.

hameerabbasi:
Probably a Numba bug. ☹️ We don't have any non-python code in Sparse.

peterbell10:
It is actually in NumPy. Reported upstream in numpy/numpy#13868

hameerabbasi:
@peterbell10 Is it possible to get a nice reproducer for the Dask issue?

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,491 @@
...
+				    &replacer,
+				    &backends,
+				    &PyTuple_Type, &def_args,
+				    &PyDict_Type, &def_kwargs,
+				    &def_impl))
+			{
+				return -1;
+			}
+
+			if (!PyCallable_Check(extractor) || !PyCallable_Check(replacer))
```
hameerabbasi: Can we make the replacer optional as well? Only time it's needed is `__ua_convert__`.

peterbell10: If there is no replacer, wouldn't that break all backends that define `__ua_convert__`?  Remembering that these would likely be defined by different people.

peterbell10:
The dask issue seems to always happen in the `sqrt` part of the call to `da.std` but if I call it directly it seems to work fine:
```python
>>> da.std(da.asarray([1, 2, 3]))
dask.array<sqrt, shape=(), dtype=float64, chunksize=()>
```

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,491 @@
...
+
+	if (def_impl_ == Py_None)
+	{
+		PyErr_SetString(
+			BackendNotImplementedError,
+			"No selected backends had an implementation for this function.");
+		return nullptr;
+	}
+
+	return PyObject_Call(def_impl_, args, kwargs);
```
peterbell10: I should probably draw attention to the fact that this is different from the current behaviour. The current behaviour requires calls to `with ua.set_backend` which would be difficult to do here (though not impossible).

uarray/_uarray_dispatch.cxx
hameerabbasi: I propose that the backend author would document the inability to support `__ua_convert__`

uarray/_uarray_dispatch.cxx
hameerabbasi: There was a reason I did that: Suppose the default implementation is in terms of other multi methods. That could be an issue.

hameerabbasi: This is hard in general: We need this current backend to be the _only backend_ for the default implementation, while allowing it to rely on other backends. The only way I found is to mix a skip and a coerce. I can push a test that does this...

hameerabbasi: Test in 604446ab904f5034c55dc85f10a2c2a6a5d1f155.

uarray/_uarray_dispatch.cxx
peterbell10: Missed your latest comment, still can't see what the issue is though. 
I check the pairs as if from `zip(args, def_args_)` starting from the end `i = len(args)-1` through to and including the first argument `i = 0`.

hameerabbasi:
@peterbell10 Would you like some cooperative development here? I'm not sure I'm good enough at C++ but I'd certainly like to try.

peterbell10:
Sure, what part did you want to work on?

uarray/_python_support.h
```diff
@@ -0,0 +1,302 @@
+
+
+#include <Python.h>
+#include "_py_ref.h"
```
hameerabbasi: I assume this line needs deleting?

hameerabbasi: Or `py_ref` needs to be moved elsewhere.

uarray/_uarray_dispatch.cxx
hameerabbasi: My mind kept going back and forth on the indices required.

uarray/_uarray_dispatch.cxx
```diff
@@ -297,29 +298,39 @@ bool for_each_backend(const std::string & domain_key, Callback call)
...
+      return ret;
   }
 
   auto & globals = global_domain_map[domain_key];
 
-  if (globals.global && !call(globals.global.get(), false))
-    return false;
+  if (globals.global)
+  {
+    ret = call(globals.global.get(), false);
```
hameerabbasi: There should be a `should_skip` here.

peterbell10: Thanks, fixed

peterbell10:
@hameerabbasi is their any chance you could look into the XND backend failures? I can't install it on my machine to debug.

hameerabbasi:
@peterbell10 Sure.

hameerabbasi:
It doesn't happen on macOS, for some reason. Let me reboot into Windows to check.

hameerabbasi:
So, it's a version issue, it's an issue with Xnd. It's happened a few times. I'd just mark the backend as `xfail` for now.

hameerabbasi:
My bad... It seems two `set_backend`s don't work inside each other and cause an infinite recursion.

hameerabbasi:
The second `set_backend` inside `xnd_backend` causes it to come back into Xnd. The innermost `set_backend` should dominate, not the outermost.

hameerabbasi:
You can mark this PR as ready-for-review and I'll give it a more thorough review. 😄

peterbell10:
In my benchmarks, the overhead is now down to ~700ns.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,1044 @@
...
+};
+
+
+struct SetBackendContext
+{
+  PyObject_HEAD
+
+  context_helper<backend_options> ctx_;
+
+  static void dealloc(SetBackendContext * self)
```
hameerabbasi: These should probably be removed. Consider the following:

```
a = ua.set_backend(be)

with a:
   # Code 1

with a:
   # Code 2
```

hameerabbasi: Ah, never mind. You're only using it in the `destructor` slot.

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,1044 @@
...
+  static PyObject * exit__(SetBackendContext * self, PyObject * /*args*/)
+    {
+      if (!self->ctx_.exit())
+        return nullptr;
+      Py_RETURN_NONE;
+    }
+};
+
+
+struct SkipBackendContext
```
hameerabbasi: Would it make the code simpler to have an abstract class and reduce the duplication?

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,1044 @@
...
+{
+  PyObject_HEAD
+
+  context_helper<py_ref> ctx_;
```
hameerabbasi: Maybe this should be the base class?

uarray/_uarray_dispatch.cxx
```diff
@@ -0,0 +1,1044 @@
...
+      if (!self->ctx_.exit())
+        return nullptr;
+      Py_RETURN_NONE;
+    }
+};
+
+
+struct SkipBackendContext
+{
+  PyObject_HEAD
```
hameerabbasi: This would be moved into the base class.

hameerabbasi:
Fantastic job! Couple of minor comments. Besides that, it might be nice to move `ua.Dispatchable` to C++, the attribute getters and the constructor might be a lot faster.

uarray/_uarray_dispatch.cxx
peterbell10: I use `context_helper` on line 735 without creating any python objects.

hameerabbasi: Fair enough, still would be nice to have a base class.

uarray/_uarray_dispatch.cxx
peterbell10: `context_helper` already covers most of the duplication. `SetBackendContext` and `SkipBackendContext` have slightly different arguments and data types so do need to be distinct.

I could maybe add something like a `py_object` base class that covers `new`, `dealloc` and holds `PyObject_HEAD`. Does that sound useful?

peterbell10:
> it might be nice to move `ua.Dispatchable` to C++

Is it alright if I leave that for a follow-up PR?

hameerabbasi:
> Is it alright if I leave that for a follow-up PR?

Of course!

uarray/_uarray_dispatch.cxx
peterbell10: I tried creating a base class and while it did seem to work, I think it's probably a bad idea to try and mix C++ inheritance with the Python c-style "inheritance". It makes very low level assumptions about the memory layout of types that don't necessarly hold for C++ types. I also get this warning that because of it:
```
uarray/_uarray_dispatch.cxx:846:12: warning: offsetof within non-standard-layout type ‘{anonymous}::Function’ is undefined [-Winvalid-offsetof]
   offsetof(Function, dict_),      /* tp_dictoffset */
```

PR #172, opened 2019-07-06
Move Dispatchable to C++.

hameerabbasi:
To do:

- [ ] Fix segfault
- [ ] Update docstrings

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -353,6 +353,85 @@ struct SetBackendContext
     }
 };
 
+struct Dispatchable
+{
+  PyObject_HEAD
+
+  py_ref value;
+  py_ref type;
+  bool coercible;
+
+  static void dealloc(Dispatchable * self)
+  {
+    PyObject_GC_UnTrack(self);
+    self->~Dispatchable();
+    Py_TYPE(self)->tp_free(self);
+  }
+
+  static PyObject * get_value(Dispatchable * self, void *)
+  {
+    PyObject * value = self->value.get();
+    Py_INCREF(value);
+    return value;
+  }
+
+  static PyObject * get_type(Dispatchable * self, void *)
+  {
+    PyObject * type = self->type.get();
+    Py_INCREF(type);
+    return type;
+  }
+
+  static PyObject * get_coercible(Dispatchable * self, void *)
+  {
+    return PyBool_FromLong(self->coercible);
+  }
+
+  static PyObject * new_(PyTypeObject * type, PyObject * args, PyObject * kwargs)
+  {
+    auto self = reinterpret_cast<Dispatchable *>(type->tp_alloc(type, 0));
+    if (self == nullptr)
+      return nullptr;
+
+    // Placement new
+    self = new (self) Dispatchable;
+    return reinterpret_cast<PyObject *>(self);
+  }
+
+  static int init(
+    Dispatchable * self, PyObject * args, PyObject * kwargs)
+  {
+    static const char *kwlist[] = {"value", "dispatch_type", "coercible", nullptr};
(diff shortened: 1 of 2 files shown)


hameerabbasi:
cc @peterbell10 It'd be nice of you to review.

uarray/_uarray_dispatch.cxx
```diff
@@ -353,6 +353,80 @@ struct SetBackendContext
...
+
+  static void dealloc(Dispatchable * self)
+  {
+    self->~Dispatchable();
+    Py_TYPE(self)->tp_free(self);
+  }
+
+  static PyObject * get_value(Dispatchable * self, void *)
+  {
+    return py_ref::ref(self->value).get();
```
peterbell10: This creates a temporary `py_ref` that is almost immediately destroyed, so the reference count doesn't actually change. You could either: use `release` instead of `get` to indicate ownership transfer; or just explicily `INCREF`
```cpp
Py_INCREF(self->value);
return self->value;
```

uarray/_uarray_dispatch.cxx
```diff
@@ -353,6 +353,80 @@ struct SetBackendContext
...
+  }
+
+  static PyObject * get_type(Dispatchable * self, void *)
+  {
+    return py_ref::ref(self->type).get();
```
peterbell10: Same here

uarray/_uarray_dispatch.cxx
```diff
@@ -353,6 +353,80 @@ struct SetBackendContext
...
+    Py_VISIT(self->type);
+    return 0;
+  }
+
+  static PyObject * repr(Dispatchable * self)
+  {
+    return PyUnicode_FromFormat("<uarray Dispatchable, value=%s, type=%s, coercible=%s>",
+      PyObject_Repr(self->value.get()),
+      PyObject_Repr(self->type.get()),
+      PyObject_Repr(py_ref::ref(PyBool_FromLong(self->coercible)).get())
```
peterbell10: https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromFormat

Use `%R` to format the repr from a `PyObject *`.

uarray/_uarray_dispatch.cxx
```diff
@@ -870,6 +944,56 @@ PyTypeObject FunctionType = {
...
+  PyVarObject_HEAD_INIT(NULL, 0)
+  "_uarray.Dispatchable",         /* tp_name */
+  sizeof(Dispatchable),                /* tp_basicsize */
+  0,                              /* tp_itemsize */
+  (destructor)Dispatchable::dealloc,  /* tp_dealloc */
+  0,                              /* tp_print */
+  0,                              /* tp_getattr */
+  0,                              /* tp_setattr */
+  0,                              /* tp_reserved */
+  (reprfunc)Function_repr,        /* tp_repr */
```
```suggestion
  (reprfunc)Dispatchable::repr,   /* tp_repr */
```

uarray/_uarray_dispatch.cxx
```diff
@@ -353,6 +353,80 @@ struct SetBackendContext
...
+{
+  PyObject_HEAD
+
+  py_ref value;
+  py_ref type;
+  bool coercible;
+
+  static void dealloc(Dispatchable * self)
+  {
+    self->~Dispatchable();
```
peterbell10: You should call `PyObject_GC_UnTrack(self)` first.

uarray/_uarray_dispatch.cxx
peterbell10: This is probably the source of the pytest segfault.

hameerabbasi:
I measured this and it seems there's no improvement at all, unfortunately.

hameerabbasi:
Thanks, @peterbell10 for the review!

PR #174, opened 2019-07-09
Improvements to _uarray.Function

peterbell10:
This PR adds to `Function`:
- `__getstate__` and `__setstate__`, allowing multimethods to be pickled.
- `tp_clear` which is used by the GC when handling cyclic references.

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -918,7 +918,7 @@ PyMethodDef Function_methods[] =
 
 PyTypeObject FunctionType = {
   PyVarObject_HEAD_INIT(NULL, 0)
-  "uarray._uarray.Function",       /* tp_name */
+  "uarray._Function",              /* tp_name */
   sizeof(Function),                /* tp_basicsize */
   0,                               /* tp_itemsize */
   (destructor)Function::dealloc,   /* tp_dealloc */
@@ -967,7 +967,7 @@ PyMethodDef SetBackendContext_Methods[] = {
 
 PyTypeObject SetBackendContextType = {
   PyVarObject_HEAD_INIT(NULL, 0)
-  "uarray._uarray.SetBackendContext",      /* tp_name */
+  "uarray._SetBackendContext",             /* tp_name */
   sizeof(SetBackendContext),               /* tp_basicsize */
   0,                                       /* tp_itemsize */
   (destructor)SetBackendContext::dealloc,  /* tp_dealloc */
@@ -1015,7 +1015,7 @@ PyMethodDef SkipBackendContext_Methods[] = {
 
 PyTypeObject SkipBackendContextType = {
   PyVarObject_HEAD_INIT(NULL, 0)
-  "uarray._uarray.SkipBackendContext",      /* tp_name */
+  "uarray._SkipBackendContext",             /* tp_name */
   sizeof(SkipBackendContext),               /* tp_basicsize */
   0,                                        /* tp_itemsize */
   (destructor)SkipBackendContext::dealloc,  /* tp_dealloc */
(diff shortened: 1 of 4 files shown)


uarray/_uarray_dispatch.cxx
```diff
@@ -822,15 +822,96 @@ int Function_traverse(Function * self, visitproc visit, void * arg)
...
+{
+  {"__getstate__", (binaryfunc)Function___getstate__, METH_NOARGS, nullptr},
+  {"__setstate__", (binaryfunc)Function___setstate__, METH_VARARGS, nullptr},
+  {NULL} /* Sentinel */
+};
+
 PyTypeObject FunctionType = {
   PyVarObject_HEAD_INIT(NULL, 0)
-  "_uarray.Function",             /* tp_name */
+  "uarray._uarray.Function",      /* tp_name */
```
hameerabbasi: I feel that the name of a private submodule shouldn't be exposed, NumPy does the same. Is it possible to change all these to `uarray.*` instead of `uarray._uarray.*`?

uarray/_uarray_dispatch.cxx
```diff
@@ -852,12 +933,12 @@ PyTypeObject FunctionType = {
    | Py_TPFLAGS_HAVE_GC),         /* tp_flags */
   0,                              /* tp_doc */
   (traverseproc)Function_traverse,/* tp_traverse */
-  0,                              /* tp_clear */
+  (inquiry)Function_clear,        /* tp_clear */
```
hameerabbasi: Is it possible to move the `Function_*` methods to be static `Function::*` methods?

hameerabbasi: (I'm more of an OOP person myself. 😄)

peterbell10:
CI is broken?
```
========================== Starting Command Output ===========================
[command]/bin/bash --noprofile --norc /home/vsts/work/_temp/24d08577-9e7e-43b0-a390-f1c033c64d10.sh
Collecting package metadata: ...working... done
Solving environment: ...working... failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - pytest-cov
Use "conda search <package> --info" to see the dependencies for each package.
```

hameerabbasi:
Let me try to create the environment locally.

hameerabbasi:
It succeeds for me. Let me try on Linux.

uarray/_uarray_dispatch.cxx
peterbell10: Do you want `uarray._Function` or do you think the type should be public?

hameerabbasi:
I could replicate the CI failure on my Linux box momentarily, but then it fixed itself. Probably an issue with conda.

uarray/_uarray_dispatch.cxx
hameerabbasi: Private. The default args/kwargs are passed into `__init__` so that could make things complicated for the user and misuse easy.

hameerabbasi:
This raises the question: Do we want to allow pickling? My view of it is that this is run-time code, and shouldn't be pickled at all. It's up to you though.

The resolution of cycles is good in case of a function with a recursive default or something along those lines, so that's understandable.

peterbell10:
The function needs to be pickleable to use with multiprocessing
https://ci.appveyor.com/project/scipy/scipy/builds/25847866

```
    @pytest.mark.parametrize("func", [fft.fft, fft.ifft, fft.rfft, fft.irfft])
    def test_multiprocess(func):
        # Test that fft still works after fork (gh-10422)
        with multiprocessing.Pool(2) as p:
>           res = p.map(func, [np.ones(100) for _ in range(4)])
func       = <uarray multimethod 'irfft'>
p          = <multiprocessing.pool.Pool object at 0x000000B2E13B9A58>
[...]
E       TypeError: can't pickle _uarray.Function objects
```

hameerabbasi:
Feel free to push the "merge" button yourself once the CI passes!

uarray/_uarray_dispatch.cxx
```diff
@@ -822,52 +829,133 @@ int Function_traverse(Function * self, visitproc visit, void * arg)
...
+
+  auto init_args = py_ref::steal(PyTuple_GetSlice(state_tuple, 0, 6));
+  if (!init_args)
+    return nullptr;
+
+  if (Function::init(self, init_args, nullptr) < 0)
+    return nullptr;
+
+  auto dict = py_ref::ref(PyTuple_GET_ITEM(state_tuple, 6));
+  if (dict == Py_None)
```
hameerabbasi: Does the test hit both branches here?

uarray/tests/test_uarray.py
```diff
@@ -28,3 +28,26 @@ def be2_ua_func(f, a, kw):
...
+
+def _replacer(args, kwargs, dispatchables):
+    return (args, kwargs)
+
+
+def test_pickle_support():
+    import pickle
+
+    mm = ua.generate_multimethod(_extractor, _replacer, "ua_tests")
+    mm.attr = "hello"
```
hameerabbasi: Do we need to allow setting attributes?

uarray/_uarray_dispatch.cxx
peterbell10: I don't think this can actually happen in normal usage because `generate_multimethod` calls `functools.update_wrapper` which adds attributes.

uarray/tests/test_uarray.py
peterbell10: Yes, mainly for things like `__name__`, `__doc__`.

uarray/_uarray_dispatch.cxx
peterbell10: I could add a test case that creates the `_Function` object directly though.

hameerabbasi: Fair enough... Would it be worth it to add a test accessing the private class or removing the branch entirely?

hameerabbasi: > I could add a test case that creates the _Function object directly though.

That’d be nice.

peterbell10: Done

PR #175, opened 2019-07-10
Retrieve state based on __module__ and __qualname__.

hameerabbasi:
Fix pickling of wrapped functions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ parse_requires()
 with open("README.md") as f:
     long_desc = f.read()
 
+
 class build_cpp11_ext(build_ext):
     def build_extension(self, ext):
         cc = self.compiler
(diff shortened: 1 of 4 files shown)


peterbell10:
I think overriding `_Function.__new__` in such a way is quite weird. Your pickling code no longer needs access to the object's internals so IMO the pickling code should be kept completely seperate.  Here's a version that works the same as yours but just from python:

```python
def pickle_function(func):
    mod_name = func.__module__
    qname = func.__qualname__

	test = unpickle_function(mod_name, qname)
	
    if test is not func:
        from pickle import PicklingError

        raise PicklingError("Can't pickle {}: it's not the same object as {}"
                            .format(func, test))

    return unpickle_function, (mod_name, qname)


def unpickle_function(mod_name, qname):
    import importlib

    try:
        module = importlib.import_module(mod_name)
        func = getattr(module, qname)
        return func
    except (ImportError, AttributeError) as e:
        from pickle import UnpicklingError

        raise UnpicklingError from e

import copyreg

copyreg.pickle(_Function, pickle_function)

```

hameerabbasi:
Superseded by #176.

PR #177, opened 2019-07-10
Remove f-strings to be Py35 compatible

peterbell10:
As you pointed out in https://github.com/scipy/scipy/pull/10383#issuecomment-509796258, f-strings cannot be used with older versions of python.

It would also be good to add a new CI check for python 3.5. I'm not familiar with azure though.

### DIFF
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -287,8 +287,8 @@ class Dispatchable:
         return (self.type, self.value)[index]
 
     def __str__(self):
-        return (
-            f"<{type(self).__name__}: type={repr(self.type)}, value={repr(self.value)}>"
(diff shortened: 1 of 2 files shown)


hameerabbasi:
I took a stab at this in #178. It seems there are unsatisfiable dependencies. This will continue to be an issue until the backends are moved out of `uarray` itself.

hameerabbasi:
Another solution is to make `unumpy` a separate repo.

peterbell10:
Could you run a reduced set of tests for the older python versions? Perhaps testing `uarray` on its own and `unumpy` with the `numpy` backend; leaving the other backends as 3.7 only.

PR #178, opened 2019-07-11
Comprehensive CI.

hameerabbasi:


hameerabbasi:
cc @rgommers would you happen to know how to fix the Windows/macOS failures? https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/anaconda?view=azure-devops&tabs=macos

rgommers:
The macOS one looks straightforward; the combination of Dask 1.2 and Python 3.5 is not supported. So simply keep 3.5 testing for Linux, and bump macOS to 3.7

The Windows one is related to default conda installs not being on any path that Windows recognizes. I can think of various fixes, but the easiest is probably to find an existing CI that uses conda on Windows (e.g. https://github.com/geopandas/geopandas/blob/master/appveyor.yml#L30)

larsoner:
To make your life easy with `conda` I'd use `ci-helpers`, it cuts down on a lot of the boilerplate code and is well maintained (has been for a few years now):

https://github.com/astropy/ci-helpers
https://github.com/mne-tools/mne-python/blob/maint/0.18/azure-pipelines.yml#L59-L74
https://github.com/mne-tools/mne-python/blob/maint/0.18/.travis.yml#L64-L65

Easily controllable using env vars:

https://github.com/mne-tools/mne-python/blob/maint/0.18/.travis.yml#L20-L26

On Anaconda, don't try to use an `environment.yml` with Windows + Python >= 3.7 though (has been broken for 6 mo with no resolution in sight):

https://github.com/conda/conda/issues/8187

larsoner:
... and beware following `appveyor.yml` code to construct your `azure-pipelines.yml`, you need to do some magical stuff at the end of commands in Azure to get env vars to persist that are not required by appveyor .

ci/azure-steps-windows.yml
```diff
@@ -0,0 +1,24 @@
...
+- script: |
+    echo "##vso[task.prependpath]$CONDA/bin"
+  displayName: Prepare conda
+
+- script: |
+    conda env update -n uarray -f .conda/environment$(python_version).yml
+  displayName: Create environment.
+
+- script: |
+    source activate uarray
```
dhirschfeld: With recent `conda` this should be:
```suggestion
    conda activate uarray
```
On Windows I'd recommend using the latest 4.7.6 as there have been various issues with earlier versions in the 4.7* series.

ci/azure-steps-windows.yml
```diff
@@ -0,0 +1,24 @@
+steps:
+- script: |
+    echo "##vso[task.prependpath]$CONDA/bin"
+  displayName: Prepare conda
```
dhirschfeld: It doesn't look like this task did anything?

image


...though maybe you only get output if you run with the debug variable set

dhirschfeld: I think perhaps it's the `$CONDA` which might not work as it's not valid `cmd.exe` syntax? It might be worth trying the pipelines syntax `$(CONDA)/bin` or `cmd.exe` syntax `%CONDA%/bin`

hameerabbasi:
Do the current failures make sense? 😕

peterbell10:
I think `black` might be the issue, from its readme:
>Black can be installed by running `pip install black`. It requires Python 3.6.0+ to run

You should only need to check code formatting on one platform anyway.

hameerabbasi:
@peterbell10 According to the error message, it seems `pip` only works on Python 3.6+. I really do think it's a solver error.

.conda/environment3.5.yml
```diff
@@ -0,0 +1,17 @@
...
+  - python=3.5
+  - pip
+  - sphinx
+  - sphinx_rtd_theme
+  - pytest
+  - pytest-cov
+  - mypy
+  - pytorch-cpu
+  - doc8
+  - black
```
peterbell10: You're still installing `black` on python35

peterbell10:
Try `pip==18.0`? Looks like that was the last version that had files with `py35` in its name on conda-forge.

https://anaconda.org/conda-forge/pip/files?version=18.0

peterbell10:
This really doesn't make sense now.
>UnsatisfiableError: The following specifications were found to be incompatible with each other:
>  - pip=18.0 -> python[version='>=3.5,<3.6.0a0']

peterbell10:
If you remove `pytest-mypy` as well, you shouldn't have to install pip right?

hameerabbasi:
This is weird, type hints existed in Py35.

ci/azure-steps-windows.yml
```diff
@@ -0,0 +1,26 @@
+steps:
+- powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
+  displayName: Add conda to PATH
+
+- script: |
+    conda config --set always_yes True
+    conda update -n base conda
+    conda init
+    conda env update -n uarray -f .conda/environment$(python_version).yml
```
larsoner: FYI Windows+3.7 can fail because of this bug:

https://github.com/conda/conda/issues/8187

But if your `pip`-installed deps don't require `numpy` you might be okay

hameerabbasi: They don't.

dhirschfeld: I wouldn't depend on pip at all when using `conda` - you can avoid a whole class of bugs and bad interaction by simply submitting PRs to [`staged-recipes`](https://github.com/conda-forge/staged-recipes) to get all your deps available with `conda`

hameerabbasi: @dhirschfeld The 3.5 deps don't include pip and it still fails... Because it doesn't recognize type hints.

hameerabbasi:
The Windows failure is unrelated. Merging.

PR #183, opened 2019-07-15
Add various requested features

### LINKED ISSUE #179: add only keyword to set_global_backend
from the discussion in scipy/scipy#10383

>    should uarray have an "only" keyword for set_global_backend as well?

not sure whether or not this would be commonly needed, but it would prevent falling back to other registered backends if set.

### LINKED ISSUE #181: add a method to clear registered and/or global backends
from the discussion in scipy/scipy#10383

>   should there be a mechanism to clear registered backends and/or the global backend?

hameerabbasi:
This PR adds:

* Clearing of backends (Fixes #181)
* Only keyword to `set_global_backend` (Fixes #179)
* Tests for registration of backends.

### DIFF
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -157,22 +157,25 @@ def generate_multimethod(
     return functools.update_wrapper(ua_func, argument_extractor)
 
 
-def set_backend(backend, *args, **kwargs):
+def set_backend(backend, coerce=False, only=False):
     """
-    A context manager that sets the preferred backend. Uses :obj:`BackendOptions` to create
-    the options, and sets the backend with the preferred options.
+    A context manager that sets the preferred backend.
 
     Parameters
     ----------
     backend
         The backend to set.
+    coerce
+        Whether or not to coerce to a specific backend's types. Implies ``only``.
+    only
+        Whether or not this should be the last backend to try.
 
     See Also
     --------
-    BackendOptions: The backend plus options.
     skip_backend: A context manager that allows skipping of backends.
+    set_global_backend: Set a single, global backend for a domain.
     """
-    return _SetBackendContext(backend, *args, **kwargs)
+    return _SetBackendContext(backend, coerce, only)
 
 
 def skip_backend(backend):
@@ -189,6 +192,7 @@ def skip_backend(backend):
     See Also
     --------
     set_backend: A context manager that allows setting of backends.
+    set_global_backend: Set a single, global backend for a domain.
     """
     return _SkipBackendContext(backend)
 
@@ -211,7 +215,7 @@ def get_defaults(f):
     return kw_defaults, tuple(arg_defaults), opts
 
 
-def set_global_backend(backend):
+def set_global_backend(backend, coerce=False, only=False):
     """
     This utility method replaces the default backend for permanent use. It
     will be tried in the list of backends automatically, unless the
@@ -230,8 +234,13 @@ def set_global_backend(backend):
     ----------
     backend
         The backend to register.
+
+    See Also
+    --------
+    set_backend: A context manager that allows setting of backends.
+    skip_backend: A context manager that allows skipping of backends.
     """
-    _uarray.set_global_backend(backend)
+    _uarray.set_global_backend(backend, coerce, only)
 
 
 def register_backend(backend):
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -259,6 +259,33 @@ def register_backend(backend):
     _uarray.register_backend(backend)
 
 
+def clear_backends(domain, registered=True, globals=False):
+    """
+    This utility method clears registered backends.
+
+    .. warning::
+        We caution library authors against using this function in
+        their code. We do *not* support this use-case. This function
+        is meant to be used only by users themselves.
+
+    Parameters
+    ----------
+    domain : Optional[str]
+        The domain for which to de-register backends. ``None`` means
+        de-register for all domains.
+    registered : bool
+        Whether or not to clear registered backends. See :obj:`register_backend`.
+    globals : bool
+        Whether or not to clear global backends. See :obj:`set_global_backend`.
+    
+    See Also
+    --------
+    register_backend : Register a backend globally.
+    set_global_backend : Set a global backend.
+    """
+    _uarray.clear_backends(domain, registered, globals)
+
+
 class Dispatchable:
     """
     A utility class which marks an argument with a specific dispatch type.
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -267,6 +267,10 @@ def clear_backends(domain, registered=True, globals=False):
         We caution library authors against using this function in
         their code. We do *not* support this use-case. This function
         is meant to be used only by users themselves.
+    
+    .. warning::
+        Do NOT use this method inside a multimethod call, or the
+        program is likely to crash.
 
     Parameters
     ----------
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -184,7 +184,7 @@ std::string backend_to_domain_string(PyObject * backend)
  * This must be installed in a python atexit handler. This prevents Py_DECREF
  * being called after the interpreter has already shudown.
  */
-PyObject * clear_all_globals(PyObject * /*self*/)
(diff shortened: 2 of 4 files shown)


uarray/_uarray_dispatch.cxx
```diff
@@ -51,11 +51,6 @@ class py_ref
       std::swap(other.obj_, obj_);
     }
 
-  friend void swap(py_ref & a, py_ref & b) noexcept
```
peterbell10: why?

uarray/_uarray_dispatch.cxx
```diff
@@ -455,7 +499,7 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
   LoopReturn ret = LoopReturn::Continue;
   for (int i = pref.size()-1; i >= 0; --i)
   {
-    auto options = pref[i];
+    auto& options = pref[i];
```
peterbell10: This will blow up if `call` modifies the list of preferred backends.

uarray/_uarray_dispatch.cxx
```diff
@@ -338,7 +382,7 @@ struct SetBackendContext
       return 0;
     }
 
-  static PyObject * enter__(SetBackendContext * self, PyObject * /*args*/)
+  static PyObject * enter__(SetBackendContext * self)
```
peterbell10: This is wrong, see the [docs](https://docs.python.org/3/c-api/structures.html#c.PyCFunction)
> `PyCFunction`
    Type of the functions used to implement most Python callables in C. Functions of this type take two `PyObject*` parameters and return one such value.

uarray/_uarray_dispatch.cxx
```diff
@@ -199,24 +193,34 @@ PyObject * clear_all_globals(PyObject * /*self*/, PyObject * /*args*/)
...
+  options.backend = py_ref::ref(backend);
+  options.coerce = coerce;
+  options.only = only;
+
+  global_domain_map[domain].global = options;
   Py_RETURN_NONE;
 }
 
-PyObject * register_backend(PyObject * /*self*/, PyObject * args)
+PyObject * register_backend(PyObject * Py_UNUSED(self), PyObject * args)
```
peterbell10: `Py_UNUSED` is _maybe_ unused, commenting out the name is compile error on use.

uarray/_uarray_dispatch.cxx
```diff
@@ -227,6 +231,46 @@ PyObject * register_backend(PyObject * /*self*/, PyObject * args)
...
+    global_domain_map[domain].global.backend = py_ref::ref(nullptr);
+  }
+}
+
+PyObject * clear_backends(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  char* domain = nullptr;
+  int registered = true, global = false;
+  if (!PyArg_ParseTuple(args, "s|pp", 
+    &domain, &registered, &global))
```
peterbell10: This won't accept domains with null characters. Just take by Object and call `domain_to_string`.

hameerabbasi: Will domains have null characters?

uarray/_uarray_dispatch.cxx
hameerabbasi: IIRC this also modifies the name so using it is a compiler error.

uarray/_uarray_dispatch.cxx
peterbell10: It's an arbitrary string so there's no reason it couldn't.  Better to use `domain_to_string` everywhere either way.

uarray/_uarray_dispatch.cxx
hameerabbasi: Interesting, I was following this: https://pythonextensionpatterns.readthedocs.io/en/latest/parsing_arguments.html#no-arguments

uarray/_uarray_dispatch.cxx
hameerabbasi: I got a warning if it existed.

uarray/_uarray_dispatch.cxx
peterbell10: https://github.com/python/cpython/blob/b49858b4b7b4c9d85ef6946ad020f83e4fa1caa7/Include/pymacro.h#L98

It adds `__attribute__((unused))` which as per GCC's docs
>This attribute, attached to a variable, means that the variable is meant to be **possibly unused**. GCC will not produce a warning for this variable. 

(Emphasis mine)

peterbell10: Actually, it does modify the name...
Still no reason to use it in C++.

uarray/_uarray_dispatch.cxx
```diff
@@ -227,6 +231,46 @@ PyObject * register_backend(PyObject * /*self*/, PyObject * args)
   Py_RETURN_NONE;
 }
 
+void clear_single(std::string domain, bool registered, bool global)
```
peterbell10: `const std::string &` to avoid a copy.

uarray/_uarray_dispatch.cxx
```diff
@@ -227,6 +231,46 @@ PyObject * register_backend(PyObject * /*self*/, PyObject * args)
...
+{
+  if (global_domain_map.find(domain) == global_domain_map.end())
```
peterbell10: Save the result from the first lookup and then you don't need to do repeated lookups.

uarray/_uarray_dispatch.cxx
```diff
@@ -227,6 +231,46 @@ PyObject * register_backend(PyObject * /*self*/, PyObject * args)
...
+
+PyObject * clear_backends(PyObject * /* self */, PyObject * args)
+{
+  PyObject* domain = nullptr;
+  int registered = true, global = false;
+  if (!PyArg_ParseTuple(args, "O|pp", 
+    &domain, &registered, &global))
+    return nullptr;
+
+  if ((!domain || domain == Py_None)  && registered && global)
```
peterbell10: `domain` can't be `nullptr` here because it's a non-optional argument.

uarray/_uarray_dispatch.cxx
```diff
@@ -190,7 +184,7 @@ std::string backend_to_domain_string(PyObject * backend)
  * This must be installed in a python atexit handler. This prevents Py_DECREF
  * being called after the interpreter has already shudown.
  */
-PyObject * clear_all_globals(PyObject * /*self*/, PyObject * /*args*/)
+PyObject * clear_all_globals(PyObject * /* self */)
```
peterbell10: This one's still wrong

uarray/_uarray_dispatch.cxx
```diff
@@ -227,6 +231,48 @@ PyObject * register_backend(PyObject * /*self*/, PyObject * args)
...
+    }
+  
+  if (registered)
+  {
+    domain_globals->second.registered.clear();
+  }
+
+  if (global)
+  {
+    domain_globals->second.global.backend = py_ref::ref(nullptr);
```
peterbell10: Nit:
```suggestion
    domain_globals->second.global.backend.reset();
```

uarray/_uarray_dispatch.cxx
```diff
@@ -338,7 +384,7 @@ struct SetBackendContext
       return 0;
     }
 
-  static PyObject * enter__(SetBackendContext * self, PyObject * /*args*/)
+  static PyObject * enter__(SetBackendContext * self, PyObject * /* self */)
```
```suggestion
  static PyObject * enter__(SetBackendContext * self, PyObject * /* args */)
```

uarray/_backend.py
```diff
@@ -250,6 +259,33 @@ def register_backend(backend):
...
+        Whether or not to clear registered backends. See :obj:`register_backend`.
+    globals : bool
+        Whether or not to clear global backends. See :obj:`set_global_backend`.
+    
+    See Also
+    --------
+    register_backend : Register a backend globally.
+    set_global_backend : Set a global backend.
+    """
+    _uarray.clear_backends(domain, registered, globals)
```
peterbell10: It's probably worth mentioning that the program is likely to crash if we're already inside a uarray method call.

grlee77:
Thanks @hameerabbasi, that was pretty quick!

PR #185, opened 2019-07-15
Potentially fix read the docs builds

peterbell10:
Read the docs does not seem to allow extension modules
https://docs.readthedocs.io/en/stable/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,6 +15,22 @@
 import os
 import sys
 
+
+# Mock Extension modules
+try:
+    import uarray._uarray
+except ImportError:
+    from unittest.mock import MagicMock
+
+    class Mock(MagicMock):
+        @classmethod
+        def __getattr__(cls, name):
+            return MagicMock()
+
+    MOCK_MODULES = ["uarray._uarray"]
+    sys.modules.update((mod, Mock()) for mod in MOCK_MODULES)
+
+
 # sys.path.insert(0, os.path.abspath('.'))
 from typing import List, Dict
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
 include versioneer.py
 recursive-include uarray *.py
 recursive-include unumpy *.py
-recursive-include ulinalg *.py


docs/conf.py
```diff
@@ -15,6 +15,22 @@
 import os
 import sys
 
+
+# Mock Extension modules
+try:
+    import uarray._uarray
+except ImportError:
```
hameerabbasi: This would allow imports when something really shouldn't be imported. How about `autodoc_mock_imports = ["uarray._uarray"]` as shown [here](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports).

peterbell10: This doesn't work because `conf.py` imports `uarray`.
```
Configuration error:                                                                                                    There is a programmable error in your configuration file:
Traceback (most recent call last):
  File "/home/peter/.local/lib/python3.7/site-packages/sphinx/config.py", line 361, in eval_config_file
    execfile_(filename, namespace)
  File "/home/peter/.local/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 86, in execfile_
    exec(code, _globals)
  File "/mnt/d/git/uarray/docs/conf.py", line 23, in <module>
    from uarray import __version__  # noqa: E402
  File "/mnt/d/git/uarray/uarray/__init__.py", line 112, in <module>
    from ._backend import *
  File "/mnt/d/git/uarray/uarray/_backend.py", line 15, in <module>
    from . import _uarray  # type: ignore
ImportError: cannot import name '_uarray' from 'uarray' (/mnt/d/git/uarray/uarray/__init__.py)  
```

hameerabbasi: It's only used to get the version. How about mirroring the code here to do that:

https://github.com/Quansight-Labs/uarray/blob/9e3c1abbebb16887c60bb7d440140c66594150d1/uarray/__init__.py#L114-L117

peterbell10: You would have to move the _version.py file outside of `uarray`, otherwise it will still try to import the whole package.

hameerabbasi: I believe not, you could do `from uarray._version import get_versions` and it should only import the right package.

peterbell10: Proof:
```
  File "/mnt/d/git/uarray/docs/conf.py", line 31, in <module>
    from uarray._version import get_version
  File "/mnt/d/git/uarray/uarray/__init__.py", line 112, in <module>
    from ._backend import *
  File "/mnt/d/git/uarray/uarray/_backend.py", line 15, in <module>
    from . import _uarray  # type: ignore
ImportError: cannot import name '_uarray' from 'uarray' (/mnt/d/git/uarray/uarray/__init__.py)
```

peterbell10: I think there is actually a way to do it but it's a hack where you use `importlib` directly.

hameerabbasi: This is weird. In any case, there are two conflicting goals here:

1. Raise an `ImportError` on a failed import.
2. Get the version into the docs.

So I see two ways out:
1. Hardcode the version (-0)
2. Use the hack. (+1, since the code won't be touched)

Moving `_version.py` out might break releases and/or version numbers. If not, I'm +1 on that too.

peterbell10: What's wrong with just leaving it as-is?

hameerabbasi: It won't raise an `ImportError` on a failed import. 🙁 So we get people with failed builds/installs trying to import it and succeeding.

hameerabbasi: And then getting weird errors when trying to actually use it.

peterbell10: This only effects the doc-builds though, right?

hameerabbasi: It affects all `ImportErrors` on `uarray._uarray`. You could use something like:

```python
if 'sphinx' not in sys.modules:
    raise
```

(From [here](https://stackoverflow.com/a/20960910/774273)).

Proof (outside of sphinx):
```python
>>> from unittest.mock import MagicMock
>>> a = MagicMock()
>>> a.a
<MagicMock name='mock.a' id='140731047645312'>
>>> a.a.a
<MagicMock name='mock.a.a' id='140731047661920'>
```

hameerabbasi: Two common cases where this may happen is a failed install/build or someone getting from source and trying to import without running `setup.py`.

peterbell10:
```python
if 'sphinx' not in sys.modules:
    raise
```
Can you describe under what situation we run `docs/conf.py` that isn't part of building the docs?

>Proof (outside of sphinx):

I don't see how that's proof of anything... Try importing `uarray` without building it first. It _will_ fail with an import error.

hameerabbasi: Ah, my bad, I neglected to see this was all in `conf.py`.

PR #187, opened 2019-07-15
Define __all__ for uarray

peterbell10:
Since `__init__.py` imports `*`, it exports things like `Callable` or `atexit`.

### DIFF
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -27,6 +27,22 @@ from ._uarray import (
     _SetBackendContext,
 )
 
+__all__ = [
+    "set_backend",
+    "set_global_backend",
+    "skip_backend",
+    "register_backend",
+    "clear_backends",
+    "create_multimethod",
+    "generate_multimethod",
+    "_Function",
+    "BackendNotImplementedError",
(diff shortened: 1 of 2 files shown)


uarray/_backend.py
```diff
@@ -27,6 +27,22 @@
...
 
+__all__ = [
+    "set_backend",
+    "set_global_backend",
+    "skip_backend",
+    "register_backend",
+    "clear_backends",
+    "create_multimethod",
+    "generate_multimethod",
+    "_Function",
```
hameerabbasi: Private variables really shouldn't be in here.

hameerabbasi: If you need to access it in tests, use `# type: ignore`

peterbell10: If you do `from uarray import *` then `_Function` still isn't included. However, sometimes you really do need to access `_Function` so it helps to have it in the `uarray` module itself.

hameerabbasi: As long as it isn't in `uarray.__all__`.

hameerabbasi:
Feel free to push the button. 😄

PR #188, opened 2019-07-15
Modify __ua_domain__ docs.

hameerabbasi:


uarray/__init__.py
```diff
@@ -12,10 +12,12 @@
 
 Let's start with the simplest:
 
-``__ua_domain__`` defines the back-end *domain*. The domain consists of a string
-that, by convention, is the module for which a back-end is provided, along with
-related packages in the ecosystem. For example, the ``"numpy"`` domain may cover
-backends for NumPy itself, as well as SciPy and the numerical scikit packages.
+``__ua_domain__`` defines the back-end *domain*. The domain consists of period-
+separated string consisting of the core dependencies plus the submodule. For
+example, if a submodule ``module2.submodule`` has a dependency on ``module1``
```
peterbell10: I'm not sure if dependency is the right word here since there can only be one. Maybe a better word is "extend", as in SciPy extends NumPy.

peterbell10: You still say "core dependencies" above this line.

PR #192, opened 2019-07-28
Fix Windows CI.

hameerabbasi:


hameerabbasi:
xref microsoft/azure-pipelines-tasks#11016

hameerabbasi:
xref conda/conda#8865

PR #195, opened 2019-08-11
 Fix write-strings warning on python < 3.7 

peterbell10:
Fixes a compiler warning specific to python < 3.7:
>`uarray/_uarray_dispatch.cxx:896:1: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]`

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -889,9 +889,11 @@ int Function::clear(Function * self)
 }
 
 
+// getset takes mutable char * in python < 3.7
+static char dict__[] = "__dict__";
 PyGetSetDef Function_getset[] =
 {
-  {"__dict__", PyObject_GenericGetDict, PyObject_GenericSetDict},
+  {dict__, PyObject_GenericGetDict, PyObject_GenericSetDict},
   {NULL} /* Sentinel */
 };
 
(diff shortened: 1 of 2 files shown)


uarray/_uarray_dispatch.cxx
```diff
@@ -889,9 +889,11 @@ int Function::clear(Function * self)
 }
 
 
+// getset takes mutable char * in python < 3.7
+static char dict__[] = "__dict__";
```
hameerabbasi: Is it possible to squeeze a const somewhere in here?

peterbell10: If the buffer was const then the pointer would have to be `const char *` and it wouldn't work.

peterbell10:
Any idea what's going on with the CI failure?

hameerabbasi:
None. I queued and it failed again... Can you try with conda<4.7, 4.7+ has a whole host of issues.

You can do that by modifying the CI files in ci/

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

larsoner:
It's probably https://github.com/conda/conda/issues/9084, in which case changing this:

https://github.com/Quansight-Labs/uarray/blob/master/.conda/environment3.7.yml

To be `3.7.3` or `3.7,!=3.7.4` (not sure about this syntax?) or something similar might work around it.

peterbell10:
Thanks @larsoner, that was the issue.

PR #199, opened 2019-09-16
Catch BackendNotImplementedError raised from __ua_function__

### LINKED ISSUE #197: Don't pass through ua.BackendNotImplementedErrors
Don't pass through `ua.BackendNotImplemented` errors and instead recognize them as a backend not implementing something.

peterbell10:
Closes #197

This catches any `BackendNotImplementedError`s raised by `__ua_function__` and stores them in a list until the end of the function call process. At which point if no backends have been called, the exception objects are made into a tuple before raising the final exception object.

This gives errors such as 
> `BackendNotImplementedError: ('No selected backends had an implementation for this function.', BackendNotImplementedError('Foo'))`

Questions:
* Should the nested-exceptions be associated with their backend? e.g. the tuple could contain tuples of `(backend, exception)`.
* Should the exception list include backends that returned `NotImplemeted`?

### DIFF
--- a/uarray/tests/test_uarray.py
+++ b/uarray/tests/test_uarray.py
@@ -16,23 +16,22 @@ class Backend:
     __ua_domain__ = "ua_tests"
 
 
-def create_nullary_mm(default=None):
-    return ua.generate_multimethod(
-        lambda: (), lambda a, kw, d: (a, kw), "ua_tests", default=default
-    )
+@pytest.fixture()
+def nullary_mm():
+    return ua.generate_multimethod(lambda: (), lambda a, kw, d: (a, kw), "ua_tests")
 
 
-def test_nestedbackend():
+def test_nestedbackend(nullary_mm):
     obj = object()
     be_outer = Backend()
     be_outer.__ua_function__ = lambda f, a, kw: obj
 
-    mm1 = create_nullary_mm()
-
     def default(*a, **kw):
-        return mm1(*a, **kw)
+        return nullary_mm(*a, **kw)
 
-    mm2 = create_nullary_mm(default=default)
+    mm2 = ua.generate_multimethod(
+        lambda: (), lambda a, kw, d: (a, kw), "ua_tests", default=default
+    )
     be_inner = Backend()
 
     def be2_ua_func(f, a, kw):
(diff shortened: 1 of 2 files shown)


hameerabbasi:
> Should the exception list include backends that returned NotImplemeted?

I think not, those failed gracefully.


> Should the nested-exceptions be associated with their backend? e.g. the tuple could contain tuples of `(backend, exception)`.

That could be useful.

hameerabbasi:
Do you have benchmarks for how much slower this is in the nominal case?

peterbell10:
In theory, if `BackendNotImplemented` isn't raised anywhere then the speed difference should be within measurement error as it's essentially just one extra (non-taken) branch per loop. 


Just running scipy.fft with this PR:
```python
In [23]: %timeit scipy.fft.fft(x)
4.43 µs ± 171 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
vs. uarray master:
```python
In [15]: %timeit scipy.fft.fft(x)
4.31 µs ± 194 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
Both are within 1 sigma so hard to tell the difference.

If I add a backend that unconditionally fails then things definitely slow down but it's hard to say how much of that is from returning vs raising.
```python
class Backend:
     __ua_domain__ = 'numpy.scipy.fft'
     @staticmethod
     def __ua_function__(f, a, kw):
         return NotImplemented # on master
         raise ua.BackendNotImplementedError # on this PR
```
| Branch | 1 failure | 2 failures | 3 failures |
|--------|------|------|------|
| PR     | 4.86 | 5.42 | 5.96 |
| Master | 4.57 | 4.88 | 5.13 |

uarray/tests/test_uarray.py
```diff
@@ -16,20 +16,23 @@ class Backend:
     __ua_domain__ = "ua_tests"
 
 
+def create_nullary_mm(default=None):
```
hameerabbasi: Can you convert this to a PyTest fixture? That’s the standard way to handle things like these,

peterbell10: Done

hameerabbasi:
Feel free to merge whenever you feel this is ready.

PR #207, opened 2019-09-26
Fix error with nested context manager.

### LINKED ISSUE #206: Issues with nested context manager when using cached object
Reproducer:

```python
class Backend:
    __ua_domain__ = "ua_bug"

    @staticmethod
    @ua.wrap_single_convertor
    def __ua_convert__(value, dispatch_type, coerce):
        return value

    @staticmethod
    def __ua_function__(func, args, kwargs):
        pass

# Should just create the context manager, not set the backend
be = ua.set_backend(Backend)


with be:
    with be:
        pass
```

Error message:

```
Traceback (most recent call last):
     ...
RuntimeError: Found invalid context state while in __exit__
```

hameerabbasi:
Fixes #206

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -280,12 +280,11 @@ class context_helper
 {
   T new_backend_;
   std::vector<T> * backends_;
-  size_t enter_size_;
+  std::vector<size_t> sizes_;
 public:
 
   context_helper():
-    backends_(nullptr),
-    enter_size_(size_t(-1))
+    backends_(nullptr)
     {}
 
   bool init(std::vector<T> & backends, T new_backend)
@@ -297,8 +296,10 @@ public:
 
   bool enter()
     {
-      enter_size_ = backends_->size();
-      try { backends_->push_back(new_backend_); }
+      try {
+        backends_->push_back(new_backend_);
+        sizes_.push_back(backends_->size());
+      }
       catch(std::bad_alloc&)
       {
         PyErr_NoMemory();
@@ -309,14 +310,20 @@ public:
 
   bool exit()
     {
-      bool success = (enter_size_ + 1 == backends_->size()
-                      && backends_->back() == new_backend_);
-      if (enter_size_ < backends_->size())
-        backends_->resize(enter_size_);
-
-      if (!success)
+      bool success = true;
+      if (sizes_.size() == 0) {
+        PyErr_SetString(PyExc_SystemExit,
+                        "Found invalid context state while in __exit__");
+        return false;
+      }
+      if (backends_->size() != sizes_.back()) {
         PyErr_SetString(PyExc_RuntimeError,
                         "Found invalid context state while in __exit__");
+        success = false;
+      }
+
+      sizes_.pop_back();
+      backends_->pop_back();
       return success;
     }
 };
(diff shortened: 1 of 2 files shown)


hameerabbasi:
It seems `test_invalid` is causing the segfault, even though it does pass. The Python process tries to `decref` something too much at one point, I think.

uarray/_uarray_dispatch.cxx
```diff
@@ -309,14 +310,21 @@ class context_helper
...
-      if (!success)
+      bool success = true;
+      if (sizes_.size() == 0 || backends_->size() != sizes_.back()) {
         PyErr_SetString(PyExc_RuntimeError,
                         "Found invalid context state while in __exit__");
+        success = false;
+      }
+      try {
+        sizes_.pop_back();
+        backends_->pop_back();
```
peterbell10: `pop_back` doesn't allocate so no need for the `try`-`catch`.

uarray/_uarray_dispatch.cxx
```diff
@@ -309,14 +310,21 @@ class context_helper
...
   bool exit()
     {
-      bool success = (enter_size_ + 1 == backends_->size()
-                      && backends_->back() == new_backend_);
-      if (enter_size_ < backends_->size())
-        backends_->resize(enter_size_);
-
-      if (!success)
+      bool success = true;
+      if (sizes_.size() == 0 || backends_->size() != sizes_.back()) {
```
peterbell10: If `sizes_.size() == 0`then calling `pop_back` is undefined behaviour.

hameerabbasi: What do you suggest we do in this case? I'm inclined to raise a `SystemError`.

peterbell10: What was wrong with the way it was done before?
```
      if (enter_size_ < backends_->size())
        backends_->resize(enter_size_);

```

hameerabbasi: `enter_size_` may not be defined if `enter_` was not called.

uarray/_uarray_dispatch.cxx
```diff
@@ -309,14 +310,20 @@ class context_helper
...
-      bool success = (enter_size_ + 1 == backends_->size()
-                      && backends_->back() == new_backend_);
-      if (enter_size_ < backends_->size())
-        backends_->resize(enter_size_);
-
-      if (!success)
+      bool success = true;
+      if (sizes_.size() == 0) {
+        PyErr_SetString(PyExc_SystemExit,
+                        "Found invalid context state while in __exit__");
```
```suggestion
                        "__exit__ call has no matching __enter__");
```

uarray/_uarray_dispatch.cxx
```diff
@@ -309,14 +310,20 @@ class context_helper
...
   bool exit()
     {
-      bool success = (enter_size_ + 1 == backends_->size()
-                      && backends_->back() == new_backend_);
-      if (enter_size_ < backends_->size())
-        backends_->resize(enter_size_);
-
-      if (!success)
+      bool success = true;
+      if (sizes_.size() == 0) {
```
```suggestion
      if (sizes_.empty()) {
```

uarray/_uarray_dispatch.cxx
peterbell10: Ah of course, `sizes_.size()` was making me think of no backends, rather than no matching `__enter__` call.

PR #209, opened 2019-09-27
Move to weakref cache.

hameerabbasi:
I'm getting no performance benefits from this cache, and I have verified that the cache is hit most of the time.

### DIFF
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -1,24 +1,16 @@
-from typing import (
-    Callable,
-    Iterable,
-    Dict,
-    Tuple,
-    Any,
-    Set,
-    Optional,
-    Type,
-    Union,
-    List,
-)
+import typing
 import inspect
 import functools
 from . import _uarray  # type: ignore
 import copyreg  # type: ignore
 import atexit
 import pickle
+import weakref
 
-ArgumentExtractorType = Callable[..., Tuple["Dispatchable", ...]]
-ArgumentReplacerType = Callable[[Tuple, Dict, Tuple], Tuple[Tuple, Dict]]
+ArgumentExtractorType = typing.Callable[..., typing.Tuple["Dispatchable", ...]]
+ArgumentReplacerType = typing.Callable[
+    [typing.Tuple, typing.Dict, typing.Tuple], typing.Tuple[typing.Tuple, typing.Dict]
+]
 
 from ._uarray import (  # type: ignore
     BackendNotImplementedError,
@@ -42,6 +34,9 @@ __all__ = [
     "all_of_type",
     "mark_as",
 ]
+
+_set_backend_cache = {}
+_skip_backend_cache = weakref.WeakKeyDictionary()
 
 
 def unpickle_function(mod_name, qname):
@@ -103,7 +98,7 @@ def generate_multimethod(
     argument_extractor: ArgumentExtractorType,
     argument_replacer: ArgumentReplacerType,
     domain: str,
-    default: Optional[Callable] = None,
+    default: typing.Optional[typing.Callable] = None,
 ):
     """
     Generates a multimethod.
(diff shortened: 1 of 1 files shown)


hameerabbasi:
Note that this PR is for discussion, not for merging.

uarray/_backend.py
```diff
@@ -191,7 +186,17 @@ def set_backend(backend, coerce=False, only=False):
     skip_backend: A context manager that allows skipping of backends.
     set_global_backend: Set a single, global backend for a domain.
     """
-    return _SetBackendContext(backend, coerce, only)
+    try:
+        return _set_backend_cache[weakref.ref(backend), coerce, only]
```
peterbell10: We can avoid the need to construct the `weakref` object by implementing in C++ where we can key by `PyObject*` without holding a reference count.

uarray/_backend.py
```diff
@@ -191,7 +186,17 @@ def set_backend(backend, coerce=False, only=False):
...
+    except KeyError:
+
+        def del_attr(x):
+            del _set_backend_cache[x, coerce, only]
+
+        ref = weakref.ref(backend, del_attr)
+        ctx = _SetBackendContext(backend, coerce, only)
+        _set_backend_cache[ref, coerce, only] = ctx
```
peterbell10: Note that the context object itself holds an owning reference to the backend; so this still leaks references.  I think this is the hardest problem to solve.

hameerabbasi: If the context object exists, we need the reference to the backend anyway.

uarray/_backend.py
hameerabbasi: How do we ensure we delete when the object goes out of scope?

uarray/_backend.py
hameerabbasi: Note also that the problem we're trying to solve is to GC the context, not the backend.

uarray/_backend.py
peterbell10: Since we don't increment the reference count, we aren't preventing GC.

uarray/_backend.py
peterbell10: I think you're describing the same issue in different words? You want to remove the context from the cache when the backend is collected, but it's existence in the cache prevents it from being collected.

uarray/_backend.py
hameerabbasi: We're still keeping around references to invalid objects in a `unordered_map` or similar.

uarray/_backend.py
hameerabbasi: We could cache the three `__ua_*__` attributes and release the backend.

hameerabbasi: Maybe the solution is to put the context in the backend's `__dict__` and release all other references to it?

hameerabbasi: ie. simply have a `__ua_cache__[("skip")|("set", coerce, only)]` attribute on the backend?

uarray/_backend.py
peterbell10: You don't ever access the backend directly, only indirectly through the context. So, this doesn't matter.

hameerabbasi: It's still leaked memory.

uarray/_backend.py
peterbell10: > We could cache the three __ua_*__ attributes and release the backend.

I think this just moves the problem to the attributes.

> Maybe the solution is to put the context in the backend's __dict__ and release all other references to it?

Good idea, I think that solves both problems doesn't it.

peterbell10: i.e. the backend owning a circular reference to itself doesn't matter so we  don't need `weakref` any more.

hameerabbasi: Opened #210 to that effect.

hameerabbasi:
Closing in favour of #210.

PR #210, opened 2019-09-27
Use a cache on the backend itself to store the context.

### LINKED ISSUE #208: Improve context manager performance
As per the discussion in https://github.com/Quansight-Labs/uarray/issues/206#issuecomment-535482750, `with ua.set_backend(be)` (and all other context managers) has a cost of ~400 ns.

This could be improved by e.g. having a global cache of `SetBackendContext` objects. Such a cache would need to use weak references to avoid leaking backend instances.

hameerabbasi:
Fixes #208

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -282,6 +282,10 @@ class context_helper
   std::vector<T> * backends_;
   std::vector<size_t> sizes_;
 public:
+  const T& get_backend() const
+    {
+      return new_backend_;
+    }
 
   context_helper():
     backends_(nullptr)
@@ -404,6 +408,12 @@ struct SetBackendContext
         return nullptr;
       Py_RETURN_NONE;
     }
+
+  static int traverse(SetBackendContext * self, visitproc visit, void * arg)
+    {
+      Py_VISIT(self->ctx_.get_backend().backend.get());
+      return 0;
+    }
 };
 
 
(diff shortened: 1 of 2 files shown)


hameerabbasi:
This is still WIP, needs `tp_visit` on both context managers.

### Benchmarks
#### Case 1
```python
with ua.set_backend(Backend):
    pass
```

#### Case 2
```python
with ctx:
    pass
```

#### `master`
Case 1: 409 ns ± 7.46 ns
Case 2: 92.5 ns ± 1.02 ns

#### `ctx-cache`
Case 1: 258 ns ± 3.27 ns
Case 2: 88.4 ns ± 1.74 ns

hameerabbasi:
This is now ready for review.

uarray/_uarray_dispatch.cxx
```diff
@@ -278,10 +278,10 @@ PyObject * clear_backends(PyObject * /* self */, PyObject * args)
 template <typename T>
 class context_helper
 {
-  T new_backend_;
-  std::vector<T> * backends_;
   std::vector<size_t> sizes_;
 public:
+  T new_backend_;
```
peterbell10: Can you at least make this read-only, i.e. add a member function `const T & get_backend() const;`.

uarray/_uarray_dispatch.cxx
```diff
@@ -278,10 +278,10 @@ PyObject * clear_backends(PyObject * /* self */, PyObject * args)
...
+  std::vector<T> * backends_;
```
peterbell10: There's no reason to make this public.

PR #212, opened 2019-09-28
Move to rich comparison of backends.

### LINKED ISSUE #211: Use rich comparison to compare backend objects
This is needed when (for example) skipping all in a factory of backends. For example, Dask might implement a backend factory and want to skip ALL of them with a single `skip_backend`.

hameerabbasi:
Fixes #211.

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -536,10 +536,10 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
 
   auto& globals = global_domain_map[domain_key];
   auto& global_options = globals.global;
-  int skip_current = global_options.backend.get() != nullptr ? should_skip(global_options.backend) : 1;
+  int skip_current = global_options.backend ? should_skip(global_options.backend) : 1;
   if (skip_current < 0)
     return LoopReturn::Error;
-  if (global_options.backend && !skip_current)
+  if (!skip_current)
   {
     ret = call(global_options.backend.get(), global_options.coerce);
     if (ret != LoopReturn::Continue)
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -496,11 +496,22 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
   auto & pref = locals->preferred;
 
   auto should_skip =
-    [&](PyObject * backend)
+    [&](PyObject * backend) -> int
       {
+        bool success = true;
         auto it = std::find_if(
           skip.begin(), skip.end(),
-          [&](const py_ref & be) { return be.get() == backend; });
+          [&](const py_ref & be)
+          {
+            auto result = PyObject_RichCompareBool(be.get(), backend, Py_EQ);
+            success = (result >= 0);
+            return (result != 0);
+          });
+
+        if (!success)
+        {
+          return -1;
+        }
 
         return (it != skip.end());
       };
@@ -509,7 +520,10 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
   for (int i = pref.size()-1; i >= 0; --i)
   {
     auto options = pref[i];
-    if (should_skip(options.backend))
+    int skip_current = should_skip(options.backend);
+    if (skip_current < 0)
+      return LoopReturn::Error;
+    if (skip_current)
       continue;
 
     ret = call(options.backend.get(), options.coerce);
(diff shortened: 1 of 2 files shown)


uarray/_uarray_dispatch.cxx
```diff
@@ -500,7 +500,7 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
       {
         auto it = std::find_if(
           skip.begin(), skip.end(),
-          [&](const py_ref & be) { return be.get() == backend; });
+          [&](const py_ref & be) { return PyObject_RichCompareBool(be.get(), backend, Py_EQ); });
```
hameerabbasi: Still needs some error detection here.

uarray/_uarray_dispatch.cxx
```diff
@@ -498,18 +498,36 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
...
-          [&](const py_ref & be) { return be.get() == backend; });
+          [&](const py_ref & be)
+          {
+            int result = PyObject_RichCompareBool(be.get(), backend, Py_EQ);
+            if (result < 0)
+            {
+              success = false;
+              return 0;
+            }
+            return result;
```
peterbell10: How about
```cpp
auto result = PyObject_RichCompareBool(be.get(), backend, Py_EQ);
success = (result >= 0);
return (result != 0);
```
Not only is it a bit more compact, it also breaks out of the `find_if` on error.

uarray/_uarray_dispatch.cxx
```diff
@@ -498,18 +498,36 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
...
-        return (it != skip.end());
+        return (int)(it != skip.end());
       };
 
   LoopReturn ret = LoopReturn::Continue;
   for (int i = pref.size()-1; i >= 0; --i)
   {
     auto options = pref[i];
-    if (should_skip(options.backend))
+    int skip = should_skip(options.backend);
```
peterbell10: You'll have to update the other uses of `should_skip` as well.

uarray/_uarray_dispatch.cxx
```diff
@@ -498,18 +498,36 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
   auto should_skip =
     [&](PyObject * backend)
```
```suggestion
     [&](PyObject * backend) -> int
```

uarray/_uarray_dispatch.cxx
```diff
@@ -498,18 +498,36 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
...
+            return result;
+          });
+
+        if (!success)
+        {
+          return -1;
+        }
 
-        return (it != skip.end());
+        return (int)(it != skip.end());
```
```suggestion
        return (it != skip.end());
```

uarray/_uarray_dispatch.cxx
```diff
@@ -522,7 +536,10 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
 
   auto& globals = global_domain_map[domain_key];
   auto& global_options = globals.global;
-  if (global_options.backend && !should_skip(global_options.backend))
+  int skip_current = global_options.backend.get() != nullptr ? should_skip(global_options.backend) : 1;
+  if (skip_current < 0)
+    return LoopReturn::Error;
+  if (global_options.backend && !skip_current)
```
```suggestion
  if (!skip_current)
```

peterbell10: You've already  done the `nullptr` check above.

uarray/_uarray_dispatch.cxx
```diff
@@ -522,7 +536,10 @@ LoopReturn for_each_backend(const std::string & domain_key, Callback call)
 
   auto& globals = global_domain_map[domain_key];
   auto& global_options = globals.global;
-  if (global_options.backend && !should_skip(global_options.backend))
+  int skip_current = global_options.backend.get() != nullptr ? should_skip(global_options.backend) : 1;
```
```suggestion
  int skip_current = global_options.backend ? should_skip(global_options.backend) : 1;
```

hameerabbasi:
### Benchmark

```python
with ua.set_backend(Backend):
    with ua.skip_backend(Backend2):
        %timeit mm()
```

`master`: 394 ns ± 6.04 ns
`rich-compare`: 412 ns ± 5.61 ns

peterbell10:
Is that benchmarking types that don't specialise the comparison operators?

hameerabbasi:
Yes, it is. The base case. Although I measured a backend with a simple `__eq__` method which uses the `is` operator for comparison, it doesn't seem to produce a measurable difference.

PR #214, opened 2019-10-15
Add clang-format file and update code style

peterbell10:
Confession time. The only reason that `uarray`'s C++ code has such a weird indent style is that I had my editor setup to work on [`pypocketfft`](https://gitlab.mpcdf.mpg.de/mtr/pypocketfft/blob/master/pypocketfft.cc#L50) when I was writing it...  This should hopefully be a bit less weird.

I could even make this a check in CI if you think that makes sense. It would certainly match with the `pytest-black` stuff in the python code.

### DIFF
--- a/ci/azure-run-clang-format.yml
+++ b/ci/azure-run-clang-format.yml
@@ -1,0 +1,16 @@
+steps:
+- bash: |
+    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+    sudo add-apt-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main"
+    sudo apt-get update
+    sudo apt-get install -y clang-format-8
+  displayName: Install clang-format
+- bash: |
+    clang-format-8 -i uarray/_uarray_dispatch.cxx
+    DIFF=$(git diff)
+    if [ $DIFF != "" ]; then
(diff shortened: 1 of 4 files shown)


hameerabbasi:
I think a separate CI entry that checked C++ formatting would be nice. 😄

peterbell10:
Yes, I haven't forgotten about this; I've just been busy working on a report that's due today. I should be able to add the CI check either this evening or at the weekend.

ci/azure-run-clang-format.yml
```diff
@@ -0,0 +1,11 @@
+steps:
+- bash: |
+    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+    sudo add-apt-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main"
+    sudo apt-get update
+    sudo apt-get install -y clang-format-8
+  displayName: Install clang-format
+- bash: |
+    clang-format-8 -i uarray/_uarray_dispatch.cxx
+    if [ "$(git diff)" != "" ]; then exit 1; fi
```
hameerabbasi: It'd be nice to print the diff to see what's wrong.

peterbell10:
The CI was supposed to fail on the latest commit, this is broken.  I was trying to push a fix but couln't use git because my syslog was apparently blowing up to 370 GB and I ran out of disk space.

PR #215, opened 2019-10-25
CI: Fix clang-format check

peterbell10:


peterbell10:
Interestingly,if I try to look at the azure for the first commit it shows "The build was manually deleted". However, when I looked at it earlier the clang-format check was correctly failing.

PR #218, opened 2019-12-31
Add support for pypy3

peterbell10:
Some of pypy3's C-API doesn't cast its arguments to `PyObject*` so the implicit
conversion from `py_ref` is never triggered. So, I remove the implicit conversion
operator and manually call `py_ref::get()` as appropriate.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,6 +35,11 @@ jobs:
     vmImage: 'ubuntu-16.04'
   steps:
     - template: ci/azure-run-clang-format.yml
+- job: pypy3
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+    - template: ci/azure-pypy3.yml
 
 
 trigger:
--- a/ci/azure-pypy3.yml
+++ b/ci/azure-pypy3.yml
@@ -1,0 +1,23 @@
+steps:
+- checkout: self
+  path: uarray
+- bash: |
+    echo "##vso[task.prependpath]$CONDA/bin"
+    conda create -y -n pypy3-env
+    conda install -y -n pypy3-env -c conda-forge pypy3.6=7.3.0
+  displayName: Prepare conda
+- bash: |
+    source activate pypy3-env
+    which pypy3
+    pypy3 -m ensurepip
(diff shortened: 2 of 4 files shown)


peterbell10:
So the `__init__.py` file in the tests folder seems to break `pypy -m pytest` where it can't import `_uarray`. Removing the file fixes `pypy` but also breaks the `cibuildwheel` jobs.

I seem to be finding conflicting advice: some say you [should have the file](https://stackoverflow.com/a/18971504),others say [not to include it](https://stackoverflow.com/a/41752043).

hameerabbasi:
How about moving the entire `tests` directory out?

hameerabbasi:
@mattip Have you ever experienced this before?

peterbell10:
I am wondering if the difference might be because I have to call pytest as a python module `pypy3 -m pytest [...]` instead of on its own `pytest [...]`. From the [pytest docs](http://doc.pytest.org/en/latest/usage.html):
>This is almost equivalent to invoking the command line script `pytest [...]` directly, except that calling via `python` will also add the current directory to `sys.path`.

Then you have an extra `uarray` in `sys.path` messing things up.

hameerabbasi:
I’m happy to merge this in, but I’d prefer to wait for Matti and see if he has an answer.

hameerabbasi:
Would the `pytest` script not be installed in the case of PyPy?

peterbell10:
The script should be there but I'm concerned about competing `pytest` versions in the `PATH`. Though I guess it might work since it's a clean `conda` environment.

peterbell10:
Okay, looks like the `conda` environment was enough to call it as just `pytest`. It still needed the working directory to be outside of the project though.

mattip:
I didn't know you could do `conda install -y -n pypy3-env -c conda-forge pypy3.6=7.3.0`. Does it still leave a cpython executable around?

hameerabbasi:
Not in the environment, it doesn't:

```
conda create -n pypy3-env -c conda-forge pypy3.6=7.3.0
conda activate pypy3-env
which python  # Gives system python
which python3 # Gives system python3
```

PR #220, opened 2020-01-09
Add retrieval and setting of state as an opaque object with pickling.

hameerabbasi:


uarray/_uarray_dispatch.cxx
```diff
@@ -179,14 +181,269 @@ std::string backend_to_domain_string(PyObject * backend) {
...
+
+    if (!output)
+      throw std::runtime_error("");
+
+    for (const auto & kv : input) {
+      py_ref py_key = convert_py(kv.first);
+      py_ref py_value = convert_py(kv.second);
+
+      Py_INCREF(py_key.get());
+      Py_INCREF(py_value.get());
```
peterbell10: [`PyDict_SetItem`](https://docs.python.org/3/c-api/dict.html#c.PyDict_SetItem) doesn't steal a reference so I don't think you should `Py_INCREF`.

uarray/_uarray_dispatch.cxx
```diff
@@ -438,7 +700,10 @@ struct SkipBackendContext {
   static int traverse(SkipBackendContext * self, visitproc visit, void * arg) {
     Py_VISIT(self->ctx_.get_backend().get());
     return 0;
-    return 0;
+  }
+
+  static PyObject * pickle_(SkipBackendContext * self, PyObject * /*args*/) {
+    return Py_BuildValue("(O)", self->ctx_.get_backend().get());
```
```suggestion
    return py_make_tuple(self->ctx_.get_backend()).release();
```

uarray/tests/test_uarray.py
```diff
@@ -255,3 +253,44 @@ def __eq__(self, other):
...
+    ua.register_backend(Backend())
+
+    with ua.set_backend(Backend()), ua.skip_backend(Backend()):
+        state = ua.get_state()
+
+    pstate = state._pickle()
+
+    assert pstate != ua.get_state()._pickle()
+
+    ua.set_state(state)
```
peterbell10: IMO this should be a context manager so that the local context is preserved for callers. Otherwise this could be pretty dangerous. e.g.
```py
def evil_function():
    ua.set_state(foo)
    # Call some multi methods

# Somewhere else
with ua.set_backend(bar):
    evil_function()
    # We have no idea what state we're in
    # ua.set_backend will probably fail in __exit__
```

uarray/_uarray_dispatch.cxx
```diff
@@ -179,14 +181,269 @@ std::string backend_to_domain_string(PyObject * backend) {
...
+    } catch (std::invalid_argument &) {
+      return nullptr;
+    } catch (std::bad_alloc &) {
+      PyErr_NoMemory();
+      return nullptr;
+    }
+  }
+
+  template <typename T, typename Convertor>
+  static std::vector<T> convert_iter(py_ref input, Convertor item_convertor) {
```
peterbell10: None of these `convert_` functions need to manage ref-counts for their `input` argument so I would just pass them as `PyObject *`.

uarray/_uarray_dispatch.cxx
```diff
@@ -179,14 +181,269 @@ std::string backend_to_domain_string(PyObject * backend) {
...
+        return nullptr;
+      local_state_t locals = convert_local_state(py_ref::ref(py_locals));
+      global_state_t globals = convert_global_state(py_ref::ref(py_global));
+
+      py_ref empty_tuple = py_ref::steal(PyTuple_New(0));
+      if (!empty_tuple)
+        return nullptr;
+
+      BackendState * output = reinterpret_cast<BackendState *>(
+          PyObject_Call(cls, empty_tuple.get(), NULL));
```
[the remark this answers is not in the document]
hameerabbasi: Yes, I've declared the method as `METH_CLASS`, which means that `&BackendStateType` will be passed in, in the usual case.

uarray/_uarray_dispatch.cxx
```diff
@@ -179,14 +181,269 @@ std::string backend_to_domain_string(PyObject * backend) {
...
+      if (output == nullptr)
+        return nullptr;
+
+      output->locals = locals;
+      output->globals = globals;
```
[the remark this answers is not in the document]
hameerabbasi: I didn't use it because an error in parsing Python objects could lead to invalid impartial state. I'll use `std::move`.

uarray/_uarray_dispatch.cxx
hameerabbasi: We have `vector<py_ref>` in the outputs, which is why I chose `py_ref` here. Is there an easy way to convert that won't require a copy?

uarray/_uarray_dispatch.cxx
```diff
@@ -179,14 +181,269 @@ std::string backend_to_domain_string(PyObject * backend) {
...
+  static global_backends convert_global_backends(py_ref input) {
+    PyObject *py_global, *py_registered;
+    if (!PyArg_ParseTuple(input.get(), "OO", &py_global, &py_registered))
+      throw std::invalid_argument("");
+
+    global_backends output;
+    output.global =
+        BackendState::convert_backend_options(py_ref::ref(py_global));
+    output.registered =
+        convert_iter<py_ref, decltype(BackendState::convert_backend)>(
```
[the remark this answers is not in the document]
hameerabbasi: Yes, it doesn't work here for some reason.

uarray/_uarray_dispatch.cxx
```diff
@@ -917,6 +1181,88 @@ PyObject * Function::get_replacer(Function * self) {
 }
 
 
+PyMethodDef BackendState_Methods[] = {
+    {"_pickle", (PyCFunction)BackendState::pickle_, METH_NOARGS, nullptr},
+    {"_unpickle", (PyCFunction)BackendState::unpickle_,
+     METH_VARARGS | METH_CLASS, nullptr},
```
hameerabbasi: `METH_CLASS` specified here.

peterbell10:
It doesn't seem to be a ref-counting issue after all. The global backend is `nullptr` by default and you can't have a `nullptr` in a tuple.

Not sure where `pytest-cov` and `pytest-black` come into this though.

hameerabbasi:
I think it was the doctests adding a null global backend to the list.

hameerabbasi:
@peterbell10 Are there any outstanding issues here?

uarray/_uarray_dispatch.cxx
```diff
@@ -917,6 +1173,89 @@ PyObject * Function::get_replacer(Function * self) {
...
+  if (!PyObject_IsInstance(
+          arg, reinterpret_cast<PyObject *>(&BackendStateType))) {
+    PyErr_SetString(
+        PyExc_TypeError, "state must be a uarray._BackendState object.");
+    return nullptr;
+  }
+
+  BackendState * state = reinterpret_cast<BackendState *>(arg);
+  local_state = state->locals;
+  global_state = state->globals;
```
peterbell10: It's occurred to me that the main motivation for this was multithreading, yet this is completely thread unsafe. We are modifying the global state without locks and it's visible to all threads.

One remedy would be to only expose the thread's local state. That should be sufficient for multithreading since the global state is shared anyway. However, it still doesn't solve `dask` because we *do* rely on different global state due to multiprocessing and delayed execution.

A more complex solution might be to change the global state to be a `thread_local std::shared_ptr<global_domain_map_t>`. By default they all point to the same global state but inside of a `ua.set_state` block, the thread has its own separate context.

Perhaps what this really means is that registered backbends are a bad idea.

hameerabbasi: How about simply making `global_state` thread-local as well?

uarray/_uarray_dispatch.cxx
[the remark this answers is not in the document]
peterbell10: Works fine for me and CI is passing. Perhaps you weren't giving it the non-deducible template parameters before?

uarray/_backend.py
```diff
@@ -65,10 +73,73 @@ def pickle_function(func):
...
+def reset_state():
+    """
+    Returns a context manager that resets all state once exited.
+
+    See Also
+    --------
+    set_state
+        Context manager that sets the backend state.
+    get_state
+        Gets a state to be set by this function.
```
hameerabbasi:
```suggestion
        Gets a state to be set by this function.
        Gets a state to be set by this context manager.
```

uarray/_uarray_dispatch.cxx
peterbell10: I suppose that would work but then they aren't really global any longer.

hameerabbasi: locals are local to a context, and globals are local to a thread.

peterbell10: Okay, should probably add a note in the docs then.

peterbell10: Oh wait, this breaks SciPy as well... We use `set_global_backend` to ensure `scipy.fft.fft` defaults to the `pocketfft` implementation.

peterbell10: That happens at module import time which is a one time operation, not once per thread.

hameerabbasi: One thing we can do is just note that global registration should only happen at import.

peterbell10: That would still be an issue for imports inside function scope.

hameerabbasi: > By default they all point to the same global state but inside of a `ua.set_state` block, the thread has its own separate context.

I did this with a bit less churn... Hopefully it's correct.

uarray/_uarray_dispatch.cxx
```diff
@@ -220,17 +481,17 @@ PyObject * register_backend(PyObject * /* self */, PyObject * args) {
...
-  global_domain_map[domain].registered.push_back(py_ref::ref(backend));
+  global_state[domain].registered.push_back(py_ref::ref(backend));
   Py_RETURN_NONE;
 }
 
 void clear_single(const std::string & domain, bool registered, bool global) {
-  auto domain_globals = global_domain_map.find(domain);
-  if (domain_globals == global_domain_map.end())
+  auto domain_globals = global_state.find(domain);
+  if (domain_globals == global_state.end())
```
peterbell10: You should be using `use_local... ? local : global` everywhere, right. For that reason, it would probably be cleaner to just have a pointer to the appropriate map.

uarray/_uarray_dispatch.cxx
```diff
@@ -917,6 +1185,99 @@ PyObject * Function::get_replacer(Function * self) {
...
+  if (!PyObject_IsInstance(
+          arg, reinterpret_cast<PyObject *>(&BackendStateType))) {
+    PyErr_SetString(
+        PyExc_TypeError, "state must be a uarray._BackendState object.");
+    return nullptr;
+  }
+
+  BackendState * state = reinterpret_cast<BackendState *>(arg);
+  local_state = state->locals;
+  global_state = state->globals;
```
peterbell10: If I have multiple levels of `ua.set_state` then this is definitely wrong. You should be writing to the global state that is about to be activated
```cpp
// update use_thread_local_globals first, then

if (use_thread_local_globals) {
  thread_global_state = state->globals;
} else {
  global_state = state->globals;
}
```

Although, I would note that `global_state = state->globals` is still very dangerous in a multi-threaded environment. For the case of `with ua.set_state()` it's probably not what you want anyway. Just setting `use_thread_local_globals` to `false` should be enough to undo the actions from inside the `ua.set_state()` block.

uarray/_uarray_dispatch.cxx
```diff
@@ -113,10 +113,12 @@ struct local_backends {
   std::vector<backend_options> preferred;
 };
 
+typedef std::unordered_map<std::string, global_backends> global_state_t;
+typedef std::unordered_map<std::string, local_backends> local_state_t;
 
 static py_ref BackendNotImplementedError;
-static std::unordered_map<std::string, global_backends> global_domain_map;
-thread_local std::unordered_map<std::string, local_backends> local_domain_map;
+static global_state_t global_state;
```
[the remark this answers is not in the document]
peterbell10: You haven't answered this yet.

uarray/_uarray_dispatch.cxx
hameerabbasi: > Although, I would note that `global_state = state->globals` is still very dangerous in a multi-threaded environment.

We're not releasing the GIL anywhere in our code, so we should be fine.



> For the case of `with ua.set_state()` it's probably not what you want anyway. Just setting `use_thread_local_globals` to `false` should be enough to undo the actions from inside the `ua.set_state()` block.

This won't work for nested `ua.set_state`s.

peterbell10: >We're not releasing the GIL anywhere in our code, so we should be fine.

The problem isn't necessarily in concurrent access to the variable but with changing the state in the middle of some function in another thread. Also note that while we don't release the GIL, we can still lose it when calling other python code. So, right in the middle of the dispatch process when we call `__ua_function__` for example.

>This won't work for nested ua.set_states.

Why wouldn't it? `ua.set_state` should only change the local state and the global state can't have been touched by the current thread.

hameerabbasi: > Why wouldn't it? `ua.set_state` should only change the `thread_local` state and the shared global state can't have been touched by the current thread.

Because of nested fork-joins. It's not necessarily one level of hierarchy. In any case, I'm not writing to the global state anymore, as far as I can see.

uarray/_uarray_dispatch.cxx
```diff
@@ -917,6 +1187,98 @@ PyObject * Function::get_replacer(Function * self) {
...
+  if (!new_tuple)
+    return nullptr;
+
+  py_ref ref = py_ref::steal(PyObject_Call(
+      reinterpret_cast<PyObject *>(&BackendStateType), new_tuple.get(), NULL));
+  BackendState * output = reinterpret_cast<BackendState *>(ref.get());
+
+  output->locals = local_domain_map;
+  output->use_thread_local_globals =
+      (current_global_state == &global_domain_map);
```
```suggestion
      (current_global_state != &global_domain_map);
```

uarray/_backend.py
```diff
@@ -62,13 +79,76 @@ def pickle_function(func):
...
+    See Also
+    --------
+    set_state
+        Sets the state returned by this function.
+    """
+    return _uarray.get_state()
+
+
+@contextlib.contextmanager
+def reset_state():
```
peterbell10: I think that if you're currently using the `global_domain_map` then `reset_state` should copy it into your `thread_global_domain_map` and use that within the current context. Basically, it should be equivalent to:
```python
with ua.set_state(ua.get_state()):
    # ...
```
Which I admit looks a bit weird but otherwise we aren't resetting the global context. (I think this is why tests are failing by the way.

uarray/_uarray_dispatch.cxx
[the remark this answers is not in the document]
hameerabbasi: Reverted this.

uarray/_uarray_dispatch.cxx
```diff
@@ -179,6 +185,261 @@ std::string backend_to_domain_string(PyObject * backend) {
...
+  static py_ref convert_py(const std::vector<T> & input) {
+    py_ref output = py_ref::steal(PyList_New(input.size()));
+
+    if (!output)
+      throw std::runtime_error("");
+
+    for (size_t i = 0; i < input.size(); i++) {
+      py_ref element = convert_py(input[i]);
+      Py_INCREF(element.get());
+      PyList_SET_ITEM(output.get(), i, element.get());
```
peterbell10: Nit: Instead of `INCREF`, better to transfer ownership with `release()`:
```suggestion
      PyList_SET_ITEM(output.get(), i, element.release());
```

uarray/_uarray_dispatch.cxx
```diff
@@ -187,6 +448,7 @@ std::string backend_to_domain_string(PyObject * backend) {
  */
 PyObject * clear_all_globals(PyObject * /* self */, PyObject * /* args */) {
   global_domain_map.clear();
+  thread_local_domain_map.clear();
```
peterbell10: We shouldn't clear `thread_local` data in this function since it will only be called on one thread. This suggests that the cleanup belongs elsewhere (probably `set_state`).

hameerabbasi:
> This is a good for now though so we can see how well dask works. (I take it from your tech-share that dask is working now?)

It is, on a local cluster. Still untested on an actual cluster, but the principle should be the same, I believe.

hameerabbasi:
Thanks so much for the in-depth review @peterbell10!

PR #231, opened 2020-03-23
Remove unused mypy ignores

person142:
Enable the `warn_unused_ignores` option in `mypy.ini` to get warnings about unused ignores and remove the unused ignores that get found. They all appear to be caused by a time before `ignore_missing_imports` was enabled.

I noticed the unused ignores with the vendored version of `uarray` in SciPy, which is currently in the ratchet:

https://github.com/scipy/scipy/blob/master/mypy.ini#L581

Figured I would try to resolve the issue upstream. Quite possible that since `warn_unused_ignores` is set on the uarray side there will be a different set of errors if SciPy updates its version however.


hameerabbasi:
CI is still failing on this one.
```
=================================== FAILURES ===================================
______________________________ uarray/_backend.py ______________________________
16: error: Cannot find implementation or library stub for module named 'uarray._uarray'
16: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
___________________________ uarray/tests/conftest.py ___________________________
6: error: Cannot find implementation or library stub for module named 'pytest_black'
14: error: Cannot find implementation or library stub for module named 'pytest_mypy'
_________________________ uarray/tests/test_uarray.py __________________________
4: error: No library stub file for module 'pytest'
4: note: (Stub files are from https://github.com/python/typeshed)
_________________________________ test session _________________________________
mypy exited with status 1.
```

person142:
Looks like `pytest --mypy` was not picking up the `mypy.ini` file. Moved it into the top level of the repo to fix that. This then caused `ignore_missing_imports = True` to get picked up (it actually wasn't getting picked up before), so I disabled it. Net effect:

- `ignore_missing_imports` remains off
- `warn_unused_ignores` enabled
- A few unnecessary `# type: ignore`s removed

hameerabbasi:
Thanks, @person142.

PR #232, opened 2020-03-26
Test PyPy3 on CIBuildWheel.

hameerabbasi:


peterbell10:
Looks like `pytest-black` doesn't build with `pypy`.

hameerabbasi:
Linux failure reported upstream at joerick/cibuildwheel#299.

PR #236, opened 2020-04-12
Add try_last option to global backends

### LINKED ISSUE #235: Chage global backend to be last when try_last=True
As title, we should process the global backend last, instead of first, if `try_last=True`.

peterbell10:
Closes #235

### DIFF
--- a/uarray/_backend.py
+++ b/uarray/_backend.py
@@ -320,7 +320,7 @@ def get_defaults(f):
     return kw_defaults, tuple(arg_defaults), opts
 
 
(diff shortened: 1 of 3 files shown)


hameerabbasi:
Seems to have some formatting mistakes as reported by `clang-tidy`.

uarray/_backend.py
```diff
@@ -339,13 +339,20 @@ def set_global_backend(backend, coerce=False, only=False):
     ----------
     backend
         The backend to register.
+    coerce : bool
```
hameerabbasi, line 342: Thanks for this!

PR #238, opened 2020-04-16
Add hierarchical domain dispatch

### LINKED ISSUE #189: Heirarchical domains
As discussed in https://github.com/scipy/scipy/pull/10383/commits/7acf8ba43485a76d720f6e62d4e510a897131eff#issuecomment-510888182

The domain `a.b.c` should be treated as a root domain `a` with sub-domains `a.b` and `a.b.c`. This means that calling a function in the leaf domain (`a.b.c`) should call backends registered to any of the super-domains, walking up the tree.

I think it would also be useful to be able to disable this on a per sub-domain basis so that the end-user can have more control over the backends being called and so they can fail-fast when not calling into the backend they expect.

peterbell10:
Closes #189

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -25,9 +25,13 @@ or temporarily.
 Domain
 ------
 
-A domain is a collection or grouping of multimethods. A domain's string,
-by convention (although not by force), is the name of the module that provides
-the multimethods.
+A domain defines the hierarchical grouping of multimethods. The domain string
+is, by convention, the name of the module that provides the multimethods.
+
+Sub-domains are denoted with a separating ``.``. For example, a multimethod in
+``"numpy.fft"`` is also considered to be in the domain ``"numpy"``. When calling
+a multimethod, the backends for the most specific sub-domain are always tried first,
+followed by the next domain up the hierarchy.
 
 Dispatching
 -----------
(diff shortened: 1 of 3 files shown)


peterbell10:
Question: how should `only=True` interact with this?  If `numpy.fft` has a backend with `only=True` should I still try the `numpy` backends?

This version does.

hameerabbasi:
Could you give the reasoning behind that decision? My gut would say no, you should terminate all processing when `only=True` because otherwise `only=True` would lose its meaning.

peterbell10:
>Could you give the reasoning behind that decision?

It's just the default that falls out if you don't make a decision

hameerabbasi:
There is one other thing that needs doing here: For backends, if the domain is of the form `a,b`, it should be added to the `unordered_map` for both `a` and `b`. This would not apply to multimethods. https://github.com/Quansight-Labs/uarray/issues/189#issuecomment-611450545

peterbell10:
Can you open a new issue for that? It doesn't really fit in with the rest of that issue.

hameerabbasi:
Sure, would you like to do it in the same PR or a different one?

peterbell10:
Separate PR ideally. I can't see there being much overlap in the implementation.

PR #243, opened 2020-04-26
Allow __ua_domain__ to be a list of domain strings

### LINKED ISSUE #239: Multiple domain support for backends
This would allow backends to support multiple domains by providing a comma separated list of domains. E.g. `"a,b"` would cover multimethods that would be covered by both `"a"` and `"b"`.

Alternate: Use a list. `["a", "b"]`.

peterbell10:
Closes #239

This allows a backend to implement multiple domains by making `__ua_domain__` a list of domains:
```python
__ua_domain__ = ["domain_a", "domain_b"]
```

TODO:

- [x] More test coverage (`skip_backend`, `register_backend`, etc.)
- [x] Benchmark function call performance. (`context_helper` in particular is used for functions with default implementations)
- [x] Validate `__ua_domain__` upfront before modifying `uarray`'s state.
- [x] Update documentation

### DIFF
--- a/uarray/small_dynamic_array.h
+++ b/uarray/small_dynamic_array.h
@@ -3,12 +3,13 @@
 #include <type_traits>
 #include <new>
 #include <cstdlib>
+#include <cassert>
 
 
 /** Fixed size dynamic array with small buffer optimisation */
-template <typename T, size_t SmallCapacity = 1>
+template <typename T, ptrdiff_t SmallCapacity = 1>
 class SmallDynamicArray {
-  Py_ssize_t size_;
+  ptrdiff_t size_;
   union {
     T elements[SmallCapacity];
     T * array;
@@ -74,7 +75,7 @@ public:
   using const_iterator = const value_type *;
   using reference = value_type &;
   using const_reference = const value_type &;
-  using size_type = Py_ssize_t;
+  using size_type = ptrdiff_t;
 
   SmallDynamicArray():
     size_(0)
--- a/uarray/small_dynamic_array.h
+++ b/uarray/small_dynamic_array.h
@@ -1,9 +1,10 @@
 #pragma once
-#include <memory>
-#include <type_traits>
-#include <new>
-#include <cstdlib>
 #include <cassert>
+#include <cstddef>
+#include <cstdlib>
+#include <memory>
+#include <new>
+#include <type_traits>
 
 
 /** Fixed size dynamic array with small buffer optimisation */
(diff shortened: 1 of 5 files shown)


hameerabbasi:
Can't we just add the backend to each domain's `vector` instead of going through all this hula-hoopery of checking the `domain` each time?

peterbell10:
What do you mean?  I do add it to each domain's vector.

hameerabbasi:
Ah, yes. I see. I guess I'm just surprised by the size of the PR.

peterbell10:
Well `set_global_backend`, `register_backend`, `set_backend`, `skip_backend` and `context_helper` all need to be modified to work with lists of domains.

peterbell10:
>Benchmark function call performance. 

In an unexpected turn of events, it appears to actually be faster than before...
For the simple benchmark below, before I get 1.77 us per call on master but now I get around 1.65 us per call. @hameerabbasi could you confirm? I'm on an old machine at the moment so that might be skewing my results.

```python
import uarray as ua

class Backend: 
    __ua_domain__ = 'ua_test' 

    def __ua_function__(self, func, a, kw): 
        if func.__name__ == 'default': 
            return object() 
        return NotImplemented

@ua.create_multimethod(lambda a, kw, d: ((),{}), domain='ua_test') 
def default(): 
    pass

func = ua.generate_multimethod(lambda: (), lambda a, kw, d: (a, kw), 'ua_test', default=lambda: default())

with ua.set_backend(Backend()): 
    %timeit func()
```

hameerabbasi:
> @hameerabbasi could you confirm?

It seems about the same for me, 948 ns vs 923.

uarray/tests/test_uarray.py
```diff
@@ -379,3 +380,50 @@ def __ua_function__(self, f, a, kw):
...
+    assert_backend_active(be)
+
+    for i in range(len(mms)):
+        ua.clear_backends(mms[i].domain, globals=True)
+
+        with pytest.raises(ua.BackendNotImplementedError):
+            mms[i]()
+
+        for j in range(i + 1, len(mms)):
+            assert mms[j]() == be.ret
```
hameerabbasi, line 424:
```suggestion
            assert mms[j]() is be.ret
```

uarray/small_dynamic_array.h
```diff
@@ -0,0 +1,228 @@
+#pragma once
```
hameerabbasi, line 1: We should probably add a copyright note if this is taken from somewhere.

peterbell10: I wrote it.

PR #244, opened 2020-05-14
Add context managers for backend determination

### LINKED ISSUE #198: add a determine_backend context manager?
So, I think we need something like

```
with unumpy.determine_backend(x):
    do_stuff
```
Where `determine_backend` checks if `x` is supported by any of the registered backends. And those backends need to auto-register on import.

Does this seem like a good idea? Or are these better ones to make library code that supports some generic array type work without the user having to set a backend?

For reference: https://mail.python.org/pipermail/numpy-discussion/2019-September/079991.html discusses the need of having library code work without the user setting a backend explicitly.

peterbell10:
Closes #198

This adds two new ways to set a local backend: `determine_backend` and `determine_backend_multi`.

```python
with ua.determine_backend_multi(
    [ua.Dispatchable(x, "mark"), ua.Dispatchable(y, "mark")],
    domain="ua_examples"
):
    pass
```

And `determine_backend` is just syntax sugar so you don't need to create `ua.Dispatchable` instances when only dispatching a single argument.


uarray/tests/test_uarray.py
```diff
@@ -427,3 +427,113 @@ def assert_backend_active(backend):
...
+        with ua.determine_backend(TypeA(), mark, domain="ua_tests"):
+            assert nullary_mm() is TypeA
+
+        with ua.determine_backend(TypeB(), mark, domain="ua_tests"):
+            assert nullary_mm() is TypeB
+
+    BackendAny = DisableBackend()  # Has no __ua_convert__, so accepts anything
+    with ua.set_backend(BackendAny):
+        with ua.determine_backend(TypeB(), mark, domain="ua_tests"):
+            assert nullary_mm() is BackendAny.ret
```
hameerabbasi, line 474: I'd change this to raise an error, if possible.

uarray/_backend.py
```diff
@@ -537,3 +539,137 @@ def __ua_convert__(self, dispatchables, coerce):
...
+    TypeA
+
+    """
+    dispatchables = (Dispatchable(value, dispatch_type, coerce),)
+    backend = _uarray.determine_backend(domain, dispatchables, coerce)
+
+    return set_backend(backend, coerce=coerce, only=only)
+
+
+def determine_backend_multi(*dispatchables, domain, only=True, coerce=False):
```
hameerabbasi, line 618: Can we add `mark=` here in case one didn't pass in `ua.Dispatchable`s? And can we change the input to an explicit iterable instead of `*dispatchables`?

peterbell10:
What's the best approach to take with the examples? As written they fail the doctest but get the message across. I'm worried the amount of code necessary to get the doctest to run would obscure the point being made. Thoughts?

uarray/_backend.py
peterbell10: Did the second, not so sure about the first. A single `mark=` argument is insufficient to cover all cases. e.g. if you have dtype dispatching you need two different dispatch types.

uarray/tests/test_uarray.py
peterbell10: Done

uarray/_backend.py
hameerabbasi: So, if the `mark=` is passed, it'd automatically mark all non-dispatchables.

hameerabbasi:
> What's the best approach to take with the examples?

I'd just put stuff in a doctest fixture. https://docs.pytest.org/en/latest/doctest.html#doctest-namespace-fixture

peterbell10:
Bit confused as to what's gone wrong here. When I run the tests locally as 
```
pytest --pyargs uarray --doctest-modules
```
The doctests are passing and the `conftest.py` is clearly being used. Yet, on the CI it seem to be using the `conftest.py` to activate `black` and `mypy` but not for the doctest fixture.

hameerabbasi:
> The doctests are passing and the `conftest.py` is clearly being used. Yet, on the CI it seem to be using the `conftest.py` to activate `black` and `mypy` but not for the doctest fixture.

I would think the `conftest.py` is picked up but not the `pytest.ini`.

hameerabbasi:
Are we sure the doc tests are being run?

peterbell10:
You can see it in the `pytest` output, from line 20 of "Lint and Tests":
```
uarray/__init__.py ...                                                   [  7%]
uarray/_backend.py ........                                              [ 26%]
uarray/_version.py ..                                                    [ 30%]
uarray/conftest.py ..                                                    [ 35%]
uarray/tests/__init__.py ..                                              [ 40%]
uarray/tests/example_helpers.py ..                                       [ 45%]
uarray/tests/test_uarray.py ......................                       [ 97%]
uarray .                                                                 [100%]
```

uarray/_backend.py
hameerabbasi: It'd still be nice to see this addressed, see my edit to my previous comment on how to make it general.

hameerabbasi:
Feel free to merge when all changes you'd like are finished.

PR #247, opened 2020-08-18
Fix refcount and GC issues

peterbell10:
Should fix the segfaults in Quansight-Labs/udiff#31

Returning `selected_backend.get()` instead of `.release()` from `determine_backend` meant the refcount was decreased without dropping the reference. The net result is decreasing the refcount of the backend once for each call to `determine_backend`. Eventually it gets GC'd and the python interpreter accesses invalid memory causing the segfault.

The GC issues are unrelated but were spotted while debugging this. The context managers had `tp_traverse` defined but the GC will only use these if the type has the `Py_TPFLAGS_HAVE_GC` flag set.

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -713,6 +713,7 @@ struct SetBackendContext {
   context_helper<backend_options> ctx_;
 
   static void dealloc(SetBackendContext * self) {
+    PyObject_GC_UnTrack(self);
     self->~SetBackendContext();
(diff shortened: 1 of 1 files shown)


hameerabbasi:
Can you get rid of MyPy completely? We're mostly untyped and it's a hindrance at this point.

hameerabbasi:
Manually triggered a new job and the failed CI passes now, so merging: https://dev.azure.com/Quansight-Labs/uarray/_build/results?buildId=766&view=results

PR #253, opened 2021-01-31
Update broken link in documentation

sayandip18:
[This link](https://github.com/Quansight-Labs/uarray/blob/master/unumpy/_multimethods.py) under [Argument Replacer](https://uarray.org/en/latest/multimethod_docs.html#argument-replacer) is dead and is updated.


hameerabbasi (approved):
Thanks, @sayandip18!

PR #255, opened 2021-04-20
Accelerate multimethod-calls with Vectorcall protocol

peterbell10:
Vectorcall is a new argument passing convention in CPython that was formalized in [PEP-590](https://www.python.org/dev/peps/pep-0590/), and made fully public in Python 3.9. It has a number of speed advantages over `tp_call` which make it desirable to use in `uarray`. It's also existed in some form since long before 3.9 as CPython's internal "FastCall" convention.

This PR provides shims to make Python 3.9-style Vectorcall available in CPython >= 3.6 and in PyPy as a wrapper around `PyObject_Call`. It also uses `Vectorcall` in places where previously I had to explicitly create a new `PyTuple` object to make the call.

In a simple benchmark, this gives a noticeable speedup on all CPython versions >= 3.7 and is neutral for Python 3.6.

| Python Version | master | vectorcall |
|----------------|--------|------------|
| 3.6            | 390 ns | 390 ns     |
| 3.7            | 275 ns | 115 ns     |
| 3.8            | 280 ns | 115 ns     |
| 3.9            | 230 ns | 103 ns     |

<details>
<summary>Benchmark code</summary>

```python
from timeit import Timer
import uarray as ua

class Backend:
    __ua_domain__ = 'bench'
    def __ua_function__(self, func, args, kwargs):
        return None

mm = ua.generate_multimethod(lambda a: (a,), lambda a, kw, d: (a, kw), 'bench')
ua.register_backend(Backend())

t = Timer('mm(None)', globals=dict(mm=mm))

results = []
n, _ = t.autorange()
n //= 10
for _ in range(10):
    t.timeit(n//10)  # Warm-up run
    results.extend([t.timeit(n) for _ in range(10)])

time = min(results) / n

print(f'{time*1e9:.2f} ns uarray overhead')
```

</details>

### DIFF
--- a/uarray/vectorcall.cxx
+++ b/uarray/vectorcall.cxx
@@ -36,6 +36,7 @@ static PyObject * build_kwarg_dict(
   return dict;
 }
 #elif (PY_VERSION_HEX < 0x03090000)
+// clang-format off
 
 static int is_method_descr(PyTypeObject* descr_tp) {
     return (
@@ -139,6 +140,7 @@ static int _PyObject_GetMethod(PyObject *obj, PyObject *name, PyObject **method)
     return 0;
 }
 
+// clang-format on
 #endif /* PYPY_VERSION */
 
 
--- a/uarray/vectorcall.cxx
+++ b/uarray/vectorcall.cxx
@@ -35,14 +35,110 @@ static PyObject * build_kwarg_dict(
   }
   return dict;
 }
-#elif ((PY_VERSION_HEX >= 0x03070000) && (PY_VERSION_HEX < 0x03090000))
-#  ifdef __cplusplus
-extern "C" {
-#  endif
-extern int _PyObject_GetMethod(PyObject *, PyObject *, PyObject **);
-#  ifdef __cplusplus
+#elif (PY_VERSION_HEX < 0x03090000)
+
+static int is_method_descr(PyTypeObject* descr_tp) {
+    return (
+        (descr_tp->tp_flags & Q_Py_TPFLAGS_METHOD_DESCRIPTOR) != 0 ||
+        (descr_tp == &PyFunction_Type) ||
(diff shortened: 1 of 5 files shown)


hameerabbasi (commented):
Would it be desirable to use the RAII wrappers in `vectorcall.c`, to avoid potential memory leaks?

hameerabbasi:
Does SciPy still support Python 3.6? [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) seems to say it doesn't. I guess we can drop support for it.

peterbell10:
There's no code here that's specifically to support Python 3.6, so I think it's fine to leave support in for now.

peterbell10:
> Would it be desirable to use the RAII wrappers in vectorcall.c, to avoid potential memory leaks?

I actually wrote it in plain C because I think this might be useful elsewhere. Each function only manages at most one object, so manual refcount isn't the _worst_ here.

uarray/vectorcall.cxx
```diff
@@ -0,0 +1,235 @@
...
+   specifically for the LOAD_METHOD opcode.
+
+   Return 1 if a method is found, 0 if it's a regular attribute
+   from __dict__ or something returned by using a descriptor
+   protocol.
+
+   `method` will point to the resolved attribute or NULL.  In the
+   latter case, an error will be set.
+*/
+static int _PyObject_GetMethod(PyObject *obj, PyObject *name, PyObject **method)
```
peterbell10, line 66: Silver lining: since we have to vendor `_PyObject_GetMethod` for windows' sake, it means we get a speedup in CPython 3.6 as well now. From 390 ns, down to 350 ns.

rgommers:
Very nice!

hameerabbasi (approved):
Thanks, @peterbell10.

PR #260, opened 2021-11-17
Modernize the CI and wheel building system.

### LINKED ISSUE #259: Move to GitHub Actions?
Would it be okay to move to GitHub actions for CI and wheel builds?

hameerabbasi:
Closes #259.


.github/workflows/build.yaml
```diff
@@ -0,0 +1,95 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+          cache: 'pip'
+      - name: Install package
+        run: |
+          pip install -ve .[tests]
+          pip install scipy==1.4
```
hameerabbasi, line 90: Should we change the version here, or add other modules?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+  clang-format:
+    name: Check format of C++ files
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run clang-format style check for C/C++ programs.
+      uses: jidicula/clang-format-action@v4.3.1
+      with:
+        clang-format-version: '13'
```
hameerabbasi, line 105: Either
```suggestion
        clang-format-version: '8'
```

or editing `.clang-format` if we don't want lint issues.

hameerabbasi: I personally like the new style better.

hameerabbasi:
This is now passing local CI: https://github.com/hameerabbasi/uarray/runs/4244490498?check_suite_focus=true

.github/workflows/build.yaml
```diff
@@ -0,0 +1,106 @@
+name: Build
+
+on: [push, pull_request]
```
hameerabbasi, line 3: Do we need additional filters here?

.github/workflows/build.yaml
peterbell10: The changes LGTM as well

.github/workflows/build.yaml
peterbell10: Perhaps bump the pin to the latest minor release which is `1.7.2`. We're still just on `fft` right now as the PR adding `uarray` to `ndimage` isn't even in master yet.

hameerabbasi:
cc @grlee77 for review.

I'm no longer authorized to merge to this repository, so someone else will have to do it. 🙂

rgommers:
> I'm no longer authorized to merge to this repository, so someone else will have to do it.

Ack! Fixed, I added you back with Maintain rights.

.github/workflows/build.yaml
rgommers: Don't think so, that should do it.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,106 @@
...
+          pytest --pyargs uarray
+  
+  scipy_fft:
+    name: Run SciPy FFT tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
```
rgommers, line 85: Python 3.7 is already dropped in SciPy master. It's fine to drop for Uarray too I think, given NEP 29.

PR #271, opened 2022-02-11
Remove one ref that lead to deinitialization of BackendNotImplementedError

oxidase:
Remove one ref that lead to deinitialization of BackendNotImplementedError after a Py_FinalizeEx that leads to segmentation faults.

### DIFF
--- a/uarray/_uarray_dispatch.cxx
+++ b/uarray/_uarray_dispatch.cxx
@@ -1816,7 +1816,6 @@ extern "C" MODULE_EXPORT PyObject * PyInit__uarray(void) {
       PyExc_NotImplementedError, nullptr));
   if (!BackendNotImplementedError)
     return nullptr;
-  Py_INCREF(BackendNotImplementedError.get());
   PyModule_AddObject(
       m.get(), "BackendNotImplementedError", BackendNotImplementedError.get());
 
--- a/uarray/tests/test_uarray.py
+++ b/uarray/tests/test_uarray.py
@@ -1,5 +1,7 @@
 import uarray as ua
 import pickle
+import subprocess
+import sys
 
 import pytest  # type: ignore
 
@@ -578,3 +580,13 @@ def test_default(nullary_mm):
     with ua.set_backend(be, coerce=True), pytest.raises(ua.BackendNotImplementedError):
         mm2()
     assert num_calls[0] == 1
+
+
+def test_refcnt():
+    out, _ = subprocess.Popen([sys.executable, "-c", """
+import uarray
+import sys
+print (sys.getrefcount(uarray._uarray.BackendNotImplementedError))
+"""], stdout=subprocess.PIPE, shell=False).communicate()
+    assert int(out.decode()) == 7
+


uarray/_uarray_dispatch.cxx
```diff
@@ -1816,7 +1816,6 @@ extern "C" MODULE_EXPORT PyObject * PyInit__uarray(void) {
       PyExc_NotImplementedError, nullptr));
   if (!BackendNotImplementedError)
     return nullptr;
-  Py_INCREF(BackendNotImplementedError.get());
```
hameerabbasi, line 1819: I don't see why an extra `INCREF` would cause segfaults. Would you link to a reproducer in an issue, add a test, and mark the PR as fixing the issue?

uarray/_uarray_dispatch.cxx
```diff
@@ -1816,7 +1816,6 @@ extern "C" MODULE_EXPORT PyObject * PyInit__uarray(void) {
...
   PyModule_AddObject(
       m.get(), "BackendNotImplementedError", BackendNotImplementedError.get());
```
hameerabbasi, line 1820:
```suggestion
      m.get(), "BackendNotImplementedError", BackendNotImplementedError.release());
```

uarray/_uarray_dispatch.cxx
oxidase: I will try to reproduce. The main problem here is that an additional reference delays removal of [BackendNotImplementedError  ](https://github.com/Quansight-Labs/uarray/blob/a5c3be24/uarray/_uarray_dispatch.cxx#L131) till a destructor call that will happen when dynamic loader will unload a dynamic library and the object is not freed in `globals_free`. In case of an embedded interpreter calling `Py_XDECREF(obj_);` after a `Py_FinalizeEx` call causes a fault.

hameerabbasi: How about `importlib.reload(uarray)`?

uarray/_uarray_dispatch.cxx
hameerabbasi: Sorry, this was incorrect.

uarray/_uarray_dispatch.cxx
oxidase: Sorry I can not reproduce the issue in a test environment. I've tried to reload  the module similar to 
```
libdl = ctypes.CDLL("libdl.so")
libdl.dlclose.argtypes = [ctypes.c_void_p]
so = ctypes.PyDLL("./uarray/_uarray.cpython-38-x86_64-linux-gnu.so")
so.PyInit__uarray.argtypes = []
so.PyInit__uarray.restype = ctypes.py_object
m = so.PyInit__uarray()
err = m.BackendNotImplementedError
del m
gc.collect(2)
libdl.dlclose(so._handle)
```
in a separate Pythnon process. It works fine and I can force so module load and reload but there is no way to call  ~py_ref for  `static py_ref BackendNotImplementedError`. So all tests are similar to a reference counter check.

Calling `.release()` has another effect as removing INCREF: it sets a nullptr to the error object  and tests show multiple assertion errors like
```
E       SystemError: <built-in function determine_backend> returned NULL without setting an error
```

hameerabbasi: I would just wait for @peterbell10's approval and CI before merging.

oxidase: CI is red, but locally tests work for me. 
I will try to dig deeper why removing `Py_INCREF` line fixes crashes at exit when the interpreter is already finalized.

uarray/_uarray_dispatch.cxx
peterbell10, line 1820: This is definitely wrong. The incref is here because `PyModule_AddObject` steals the reference, so now the destructor of `BackendNotImplementedError` will decref a reference it doesn't own.

It would be helpful if you open an issue with a clear description of the issues you're running into.

peterbell10 (changes requested):

uarray/_uarray_dispatch.cxx
hameerabbasi: It's a static variable I believe, so is the destructor called for sure?

peterbell10: Yes, the destructor is called at program exit. Although strictly speaking the actual decref should happen in `globals_free`:
https://github.com/Quansight-Labs/uarray/blob/f390751af826a629995c20078ea5927eee3663cf/uarray/_uarray_dispatch.cxx#L541

hameerabbasi: Would it then happen twice according to the C++ standard, once in `globals_free` and once when the library is unloaded?

peterbell10: `reset` clears the `PyObject*` to `nullptr`, so no.

hameerabbasi:
Closing due to https://github.com/Quansight-Labs/uarray/pull/271#discussion_r805075523.

oxidase:
@hameerabbasi sorry for the noise. The issue was on my side.

I found the reason for segmentation violations that is related to the deinitialization order of global static py_ref variables and `globals_free` call that happened after destructors calls. In my case `BackendNotImplementedError.reset()` was called for an already destructed `BackendNotImplementedError` object and this lead to memory corruptions with  stack traces like
```
#0  unlink_chunk (p=p@entry=0x6ef2a0, av=0x7ffff7807b80 <main_arena>) at malloc.c:1466
#1  0x00007ffff76b600b in _int_free (av=0x7ffff7807b80 <main_arena>, p=0x6ee1e0, have_lock=<optimized out>) at malloc.c:4342
#2  0x00007ffff7a0f78c in type_dealloc (type=0x6eee70) at ../Python-3.8.12/Objects/typeobject.c:3368
#3  0x00007ffff79a373f in _Py_DECREF (filename=<synthetic pointer>, lineno=541, op=<optimized out>) at ../Python-3.8.12/Include/object.h:478
#4  _Py_XDECREF (op=<optimized out>) at ../Python-3.8.12/Include/object.h:541
#5  free_keys_object (keys=0x6ef710) at ../Python-3.8.12/Objects/dictobject.c:584
#6  0x00007ffff7a0f616 in type_clear (type=0x617290) at ../Python-3.8.12/Objects/typeobject.c:3607
#7  0x00007ffff79ee10f in delete_garbage (state=<optimized out>, state=<optimized out>, old=0x7ffff7b9b9a0 <_PyRuntime+416>, collectable=0x7fffffffbfa0) at ../Python-3.8.12/Modules/gcmodule.c:948
#8  collect (generation=2, n_collected=0x0, n_uncollectable=0x0, nofail=1, state=<optimized out>) at ../Python-3.8.12/Modules/gcmodule.c:1123
#9  0x00007ffff7a403c2 in _PyGC_CollectNoFail () at ../Python-3.8.12/Modules/gcmodule.c:1877
#10 0x00007ffff7a2ff3e in PyImport_Cleanup () at ../Python-3.8.12/Python/import.c:599
#11 0x00007ffff7a3551b in Py_FinalizeEx () at ../Python-3.8.12/Python/pylifecycle.c:1229
#12 0x000000000044e599 in py::~py() () at py.cc:88
#13 0x00007ffff7665a27 in __run_exit_handlers (status=0, listp=0x7ffff7807718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#14 0x00007ffff7665be0 in __GI_exit (status=<optimized out>) at exit.c:139
#15 0x00007ffff76430ba in __libc_start_main (main=0x4caa52 <main>, argc=3, argv=0x7fffffffc2b8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffc2a8) at ../csu/libc-start.c:342
#16 0x000000000044cbe9 in _start ()
```
where deallocated objects unrelated to `uarray`.

PR #273, opened 2022-03-01
ENH: Add annotations for the `uarray._uarray` extension module

BvB93:
Recently I noticed in scipy that uarray is still lacking type annotations, so if there is any interest I would very much like to contribute them.

This PR in particular focuses on the `uarray._uarray` extension module. As the latter is written in C++, it is necessary to resort to .pyi stub files for distributing PEP 484 type annotations.

### DIFF
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -2,3 +2,4 @@ pytest>=3.5
 pytest-flake8
 pytest-cov
 pytest-black; platform_python_implementation == 'CPython'
+mypy>=0.930
--- a/uarray/_typing.pyi
+++ b/uarray/_typing.pyi
@@ -1,0 +1,54 @@
+"""Helper module with various typing-related utilities."""
+
+from collections.abc import Callable
+from typing import Any, Protocol, TypeVar, Iterable, type_check_only
+
+import uarray
+
+_T = TypeVar("_T")
+_T_co = TypeVar("_T_co", covariant=True)
+
+@type_check_only
+class _PySequence(Protocol[_T_co]):
+    def __len__(self) -> int: ...
+    def __getitem__(self, key: int, /) -> _T_co: ...
+
+@type_check_only
+class _SupportsUADomain(Protocol):
+    @property
+    def __ua_domain__(self) -> str | _PySequence[str]: ...
+
+@type_check_only
+class _SupportsUAConvert(Protocol):
+    def __ua_convert__(
+        self,
+        dispatchables: tuple[uarray.Dispatchable, ...],
+        coerce: bool,
+        /,
+    ) -> Iterable[Any]: ...
+
+_ReplacerFunc = Callable[
+    [
+        tuple[Any, ...],
+        dict[str, Any],
+        tuple[uarray.Dispatchable, ...],
+    ],
+    tuple[tuple[Any, ...], dict[str, Any]],
+]
+
+_PyGlobalDict = dict[
+    str,
+    tuple[
+        tuple[_T | None, bool, bool],
+        list[_T],
+        bool,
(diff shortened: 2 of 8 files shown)


uarray/_typing.pyi
```diff
@@ -0,0 +1,54 @@
+"""Helper module with various typing-related utilities."""
```
BvB93, line 1: The content of this small helper module could in principle be merged with `_uarray.pyi`, but having the proper type annotations separate from helper types minimizes clutter somewhat, in my opinion.

uarray/_uarray.pyi
```diff
@@ -0,0 +1,126 @@
...
+        self,
+        extractor: Callable[..., tuple[uarray.Dispatchable, ...]],
+        replacer: None | _ReplacerFunc,
+        domain: str,
+        def_args: tuple[Any, ...],
+        def_kwargs: dict[str, Any],
+        def_impl: None | Callable[..., Any],
+    ) -> None: ...
+    def __repr__(self) -> str: ...
+    def __call__(self, *args: Any, **kwargs: Any) -> Any: ...
```
hameerabbasi, line 81: Would there be a way to make this have the "same input signature" as the `extractor`? I'm assuming not, but it's worth asking.

BvB93: It's a good thing you asked in that case, because the answer is "yes". Initially I had some trouble figuring out the signature of `__call__`, which is why I left it as broad as it is now.

hameerabbasi: Please feel free to ask if there are other things that are left unspecified. Very willing to help move this along.

uarray/_uarray.pyi
```diff
@@ -13,6 +16,9 @@ from uarray._typing import (
     _SupportsUADomain,
 )
 
+_P = ParamSpec("_P")
+_ReturnT = TypeVar("_ReturnT", bound=tuple[uarray.Dispatchable, ...])
```
hameerabbasi, line 20: I would make this `Any`, the output isn't matched.

hameerabbasi:
Could you add `mypy` to the linting stage?

uarray/_uarray.pyi
BvB93: For now I've simply removed the return-typevar. 

It could considered to reintroduce it at some point with its default set to `Any` (so users can more easily manually declare the return-type), but this might be somewhat tricky as the return type can very easily change again by selecting a different backend (via the likes of `uarray.set_backend`). Simply sticking to `Any` for now is probably the safest and easiest option.

requirements/tests.txt
```diff
@@ -2,3 +2,4 @@ pytest>=3.5
 pytest-flake8
 pytest-cov
 pytest-black; platform_python_implementation == 'CPython'
+mypy>=0.930
```
BvB93, line 5: This is the first [mypy version](https://mypy-lang.blogspot.com/2021/12/mypy-0930-released.html) with support for `ParamSpec`.

BvB93 (commented):
A minor rebase was necessary due to the changes in https://github.com/Quansight-Labs/uarray/pull/272, but the mypy linting job should now be up and running.

BvB93:
As a gentle reminder, are any other changes required to get this merged?

hameerabbasi (approved):
Sorry, @BvB93, This Fell off my radar.

LGTM. @peterbell10, would you care to review? If not, I'd merge this.

rgommers (approved):
LGTM too, in it goes. Thanks a lot @BvB93, very nice to have type annotations! And thanks @hameerabbasi for the review!

PR #275, opened 2022-07-05
CI: fix doc build warning that trips up CI job

rgommers:


rgommers:
doc build is green, so merging

hameerabbasi:
Thanks for the fix @rgommers!

PR #278, opened 2022-07-11
ENH: Add annotations for `uarray._backend`

BvB93:
Follow up on https://github.com/Quansight-Labs/uarray/pull/273

Adds annotations for all remaining functions and marks `uarray` is a fully typed package.


uarray/_backend.py
```diff
@@ -435,23 +488,36 @@ class Dispatchable:
...
-    def __getitem__(self, index):
+    @overload
+    def __getitem__(self, index: Literal[0]) -> _TT:
+        ...
+
+    @overload
+    def __getitem__(self, index: Literal[1]) -> _T:
+        ...
+
+    def __getitem__(self, index: Literal[0, 1]) -> _TT | _T:
```
BvB93, line 509: I'm not a huge fan of overloads with online annotations, but this is a situation where we really cannot do without.

uarray/_backend.py
```diff
@@ -66,7 +85,13 @@ def unpickle_function(mod_name, qname, self_):
...
 
 
-def pickle_function(func):
+@no_type_check
+def pickle_function(
+    func: Callable[..., Any],
+) -> tuple[
+    Callable[[str, str, object], Callable[..., Any]],
+    tuple[str, str, Any | None],
+]:
```
BvB93, line 94: The `try`/`except` blocks here are simply too complicated for type checkers to understand, unfortunately.

uarray/_backend.py
```diff
@@ -618,8 +703,13 @@ def determine_backend(value, dispatch_type, *, domain, only=True, coerce=False):
...
 def determine_backend_multi(
-    dispatchables, *, domain, only=True, coerce=False, **kwargs
-):
+    dispatchables: Iterable[Any],
+    *,
+    domain: str,
+    only: bool = True,
+    coerce: bool = False,
+    **kwargs: type[Any],
+) -> _SetBackendContext:
```
BvB93, line 712: You could overload here w.r.t. the `dispatch_type` keyword-only kwarg, but I'm not sure it's worth the extra `overload` clutter.

hameerabbasi:
Thanks for the PR, @BvB93. I will review on the weekend, or even before if I get the time. Feel free to ping me if you need a review earlier or if I forget.

uarray/_backend.py
```diff
@@ -175,11 +212,11 @@ def wrapper(a):
...
 
 def generate_multimethod(
-    argument_extractor: ArgumentExtractorType,
-    argument_replacer: ArgumentReplacerType,
+    argument_extractor: Callable[_P, tuple[Dispatchable[Any, Any], ...]],
+    argument_replacer: _ReplacerFunc,
     domain: str,
-    default: typing.Optional[typing.Callable] = None,
-):
+    default: None | Callable[..., Any] = None,
```
hameerabbasi, line 218: This should have the same signature as `argument_extractor`.

uarray/_backend.py
hameerabbasi: It would be incorrect, as even with a default `dispatch_type`, the output types are arbitrary.

hameerabbasi (approved):
Couple of minor comments, I'd be willing to proceed without them.

hameerabbasi:
Thanks for the PR, @BvB93. This is in.

